PR #1, opened 2017-06-11
Calendar dates issues

Chauce:
Current issues with the date selector: 
- Dates aren't being saved
- Trying to edit and then save a passenger give you this error: `undefined method `update' for nil:NilClass`


app/views/passengers/index.haml
```diff
@@ -0,0 +1,57 @@
+= form_tag(passengers_path, :method => "get") do
```
dfaulken: Best to avoid hashrockets, and you don't need those parentheses either:

```ruby
form_tag passengers_path, method: :get do
```

app/views/passengers/index.haml
```diff
@@ -0,0 +1,57 @@
...
+  = radio_button_tag :filter, 'permanent'
+  Temporary
+  = radio_button_tag :filter, 'temporary'
+  Show All
+  = radio_button_tag :filter, 'all'
+  = submit_tag "Find passengers"
+%p
+= link_to "Add New Passenger", new_passenger_path
+
+- if @permanent.present?
```
dfaulken: You coerced it to a true or false, so it's a little misleading (I think) to be checking to see if it's more generally `present?`.

app/views/passengers/index.haml
```diff
@@ -0,0 +1,57 @@
...
+        %th
+        %th
+        %th
+    %tbody
+      - @passengers.each do |passenger|
+        %tr
+          %td= passenger.name
+          %td= passenger.email
+          %td= passenger.phone
+          - if passenger.active
```
dfaulken: For boolean columns, it's nice to query them using a predicate method (here, `active?`). This means you don't have to worry about defining a third behavior for when it might be `nil`. Predicate methods coerce that to `false`.

app/views/passengers/index.haml
```diff
@@ -0,0 +1,57 @@
...
+            %td Yes
+          - else
+            %td No
```
dfaulken: This is a great case to define an application helper method, since displaying 'Yes' or 'No' based on a boolean value is something you'll be doing all the time. Say you wrote a method that used a boolean argument to return 'Yes' or 'No' and called it `bananas` — then you could write

```haml
%td= bananas passenger.active?
```

app/views/passengers/show.haml
```diff
@@ -0,0 +1,39 @@
+%p#notice= notice
+
+%h1= @passenger.name
```
dfaulken: :+1: for the convention of having your page 'title' be the name of the object rather than just 'Displaying Passenger' or something painfully obvious and unhelpful.

app/views/sessions/new.html.haml
```diff
@@ -0,0 +1,12 @@
+- provide(:title, "Log in")
```
dfaulken: Does this work? If so, how?

app/views/sessions/new.html.haml
```diff
@@ -0,0 +1,12 @@
...
+.row
+  .col-md-6.col-md-offset-3
+    = form_for(:session, url: login_path) do |f|
+      = f.label :email
+      = f.email_field :email, class: 'form-control'
+      = f.label :password
+      = f.password_field :password, class: 'form-control'
+      = f.submit "Log in", class: "btn btn-primary"
+    %p
+      New user? #{link_to "Sign up now!", login_path}
```
dfaulken: Did this come from somewhere? If so, :-1: to them for this string interpolation (rather than just a new line beginning with an `=`).

app/views/users/_user.json.jbuilder
```diff
@@ -0,0 +1,2 @@
+json.extract! user, :id, :name, :email, :phone, :spire, :active, :admin, :created_at, :updated_at
```
dfaulken: Here too with the `jbuilder`, if you don't have immediate plans to use it, it's trivial to add in later. I would just remove it now.

config/schedule.rb
```diff
@@ -0,0 +1,21 @@
+every 1.day, :at => '4:30 am' do
+  runner "Passenger.update_active"
```
dfaulken: The `update_active` method you defined is a model method, not a class method, so this definitely won't work.

Also, I think you can just write `every :day`.

Also it's nice to remove all of those comments ⬇️  once you're actually using a file for its intended purpose.

app/models/passenger.rb
```diff
@@ -0,0 +1,17 @@
...
+                    format: { with: VALID_EMAIL_REGEX }, uniqueness: true
+
+  scope :permanent, -> { where(permanent: true) }
+  scope :temporary, -> { where(permanent: false) }
+
+  def self.search(search)
+    where("name LIKE ?", "%#{search}%")
+  end
+
+  def update_active
```
dfaulken: Edit: it seems from your whenever file that maybe you were intending this method to do something other than what it does?

app/views/sessions/new.html.haml
dfaulken: If it came from you, then I don't want to be that harsh.

dfaulken:
Also, I don't know if Karin is choosing not to do this ATM to avoid throwing too many things at you, but I'd recommend using `rubocop` from the ground up. The plan for this app is likely to integrate Codeclimate soon, and Travis probably a little later, so might as well not have to make a million changes down the line (and also, adhering to good style practices makes your code that much easier to review 😄).

Chauce:
Thanks for taking a look through! I wasn't sure if that was the right way
to ask people for feedback or not; I also just wanted to keep some sort of
log of the issues I was running into in case I forgot before Monday.

I actually fixed that NilClass error; I forgot a

> @passenger = Passenger.find(params[:id])
>
line in the update def in the passengers controller. So that's squared
away.

This is especially relevant to Ava and Karin (so I hope y'all see this!) :
Dates saved once and I thought I was good so I pushed what I had and ...
then it quit working again for reasons I've yet to discern, because I
didn't change anything after the first time it worked. There's also this
weird thing where the calendar doesn't show up on the create/edit passenger
page *unless *you refresh the page. Weird. And probably related? I used the
debugger console in Firefox and didn't notice anything different between
the initial and refreshed pages, but maybe a different set of eyes can spot
something.

​As for everything else, I will attend to it in the morning!

Best,
-Chauce​

------------------------------------------------------------
---------------------------
Chasya "Chauce" Church
UMass Transit IT
Physics B.S. - UMass Amherst
Minors: Computer Science, Applied Mathematics, Native Studies
Pronouns are important! Mine are: They/them/their

app/controllers/passengers_controller.rb
```diff
@@ -61,7 +61,11 @@ def destroy
   private
 
   def passenger_params
-    params.require(:passenger).permit(:name, :address, :email, :phone,
+    #binding.pry
+    permitted_attrs = params.require(:passenger).permit(:name, :address, :email, :phone,
       :wheelchair, :active, :permanent, :expiration, :note)
+    expiration_date = permitted_attrs[:expiration]
+    permitted_attrs[:expiration] = Date.strptime(expiration_date, '%m/%d/%Y')
```
dfaulken: :-1: There should definitely be a way to get the datepicker to send data in the format you need.

werebus: https://api.jqueryui.com/datepicker/#option-dateFormat

Gemfile
```diff
@@ -33,10 +36,13 @@ gem 'jbuilder', '~> 2.5'
 
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
+gem 'haml', '5.0.0.beta.2'
```
Chauce: I don't recall adding that one ... I think it may have been in there when we started the project. Possibly added by Karin?

app/views/passengers/index.haml
Chauce: They're actually checking for the presence of an instance variable, so this is proper.

dfaulken: But you define it on line 7 of the controller. It's always 'present' (as in defined), but it's not always `present?` (as in false vs. true). That's why I'm saying it's misleading.

app/views/passengers/index.haml
Chauce: Fixed .... on a different branch. So it just won't be part of this pull request.

app/views/sessions/new.html.haml
Chauce: It does, but we're going to ditch it because we just need a developer login at the moment to test admin vs non-admin functionality.

config/schedule.rb
Chauce: All fixed in a future branch!

Chauce:
@dfaulken - Finished up all the issues except for the stuff Karin said was fine/things we handled in future branches.

dfaulken:
It's @Anbranin's call, but there are a number of them unchanged since I commented. I'd be interested to learn why those are wontfix issues (if they are).

Chauce:
Forgot to push! Just did so now.

app/controllers/passengers_controller.rb
```diff
@@ -15,56 +17,41 @@ def index
...
-      if @passenger.update(passenger_params)
-        format.html { redirect_to @passenger, notice: 'Passenger was successfully updated.' }
-        format.json { render :show, status: :ok, location: @passenger }
-      else
-        format.html { render :edit }
-        format.json { render json: @passenger.errors, status: :unprocessable_entity }
-      end
+    if @passenger.update(passenger_params)
+      redirect_to @passenger, notice: 'Passenger was successfully updated.'
+    else
```
dfaulken: Rubocop would prefer inline here:

```ruby
else render :edit
```

I definitely do recommend incorporating it sooner rather than later.

app/controllers/users_controller.rb
```diff
@@ -28,54 +28,37 @@ def create
...
-      params.require(:user).permit(:name, :email, :phone, :spire, :active, :admin)
-    end
-  def require_admin_in_roster_or_self
-    return if @current_user == @user || @current_user.admin_in?(@roster)
-    # ... and return is correct here
-    # rubocop:disable Style/AndOr
-    head :unauthorized and return
-    # rubocop:enable Style/AndOr
+  def user_params
+    params.require(:user).permit(:name, :email, :phone, :spire, :active, :admin)
```
dfaulken: If I'm not an admin, can I change whether or not I'm an admin by injecting `admin: true` while changing my email or something?

If that's out of scope now, that's fine.

app/controllers/passengers_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+        format.json { render :show, status: :ok, location: @passenger }
+      else
+        format.html { render :edit }
+        format.json { render json: @passenger.errors, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  def destroy
+    @passenger.destroy
```
dfaulken: You have done abstraction now!

PR #2, opened 2017-06-15
Color codes

avacmeyer:


app/assets/stylesheets/passengers.scss
```diff
@@ -37,16 +37,37 @@ a {
 }
 
 table{
+  border: 2px solid #999;
+  th.yesno {
+    text-align: center;
+    padding: 5px 25px 5px 25px;
```
dfaulken: Don't forget that if you want top-bottom the same, and left-right the same, you can do:

```css
padding: 5px 25px;
```

PR #3, opened 2017-06-15
Seeds

Chauce:


Chauce:
I want to figure out how to use it later, and was trying to make it work initially, but Matt suggested I seed the way that things are listed in the current seeds.rb just for ease of use at the moment.

dfaulken:
Okay, sounds good.

PR #4, opened 2017-06-20
Made color codes

avacmeyer:
screen shot 2017-06-20 at 2 47 56 pm

### DIFF
--- a/app/assets/javascripts/datatables.js
+++ b/app/assets/javascripts/datatables.js
@@ -9,5 +9,7 @@ $(document).ready( function () {$('#passengers').DataTable({
     { "bSortable": false },
     { "bSortable": false },
     { "bSortable": false },
-  ]
+  ],
+  "aLengthMenu": [[25, 50, 75, 100, -1], [25, 50, 75, 100, "All"]],
+  "iDisplayLength": 100
 });});
--- a/app/assets/javascripts/datatables.js
+++ b/app/assets/javascripts/datatables.js
@@ -1,0 +1,13 @@
+$(document).ready( function () {$('#passengers').DataTable({
+  "aoColumns": [
+    null,
+    null,
+    null,
+    { "bSortable": false },
+    { "bSortable": false },
+    null,
+    { "bSortable": false },
+    { "bSortable": false },
+    { "bSortable": false },
+  ]
+});});
--- a/app/assets/stylesheets/passengers.scss
+++ b/app/assets/stylesheets/passengers.scss
@@ -36,7 +36,7 @@ a {
   }
 }
 
-table{
+table#passengers{
   border: 2px solid #999;
   th.yesno {
     text-align: center;
--- a/app/assets/stylesheets/passengers.scss
+++ b/app/assets/stylesheets/passengers.scss
@@ -75,9 +75,6 @@ table#key, table#passengers {
 }
 
 table#key {
-  //td.space {
-  //  padding-left: 320px;
-  //}
   td {
     font-size: 13px;
     padding: 5px;
--- a/app/assets/stylesheets/passengers.scss
+++ b/app/assets/stylesheets/passengers.scss
@@ -11,12 +11,24 @@ body {
   line-height: 18px;
 }
 
-p, ol, ul, td {
+ol, ul, td {
   font-family: verdana, arial, helvetica, sans-serif;
   font-size: 13px;
   line-height: 18px;
 }
-
+b.yay {
+  font-family: verdana, arial, helvetica, sans-serif;
+  font-size: 18px;
+  line-height: 18px;
+  background-color: #e0dbfa;
+  text-decoration: underline;
+}
+p {
+  font-family: verdana, arial, helvetica, sans-serif;
+  font-size: 19px;
+  line-height: 18px;
+  color: #25da6d;
+}
 pre {
   background-color: #eee;
   padding: 10px;
@@ -33,6 +45,41 @@ a {
   &:hover {
     color: #fff;
     background-color: #000;
+  }
+  .radioDiv {
+   width:200px;
+  }
+  .leftRadio {
+   float:left;
+  }
+  .rightRadio {
+   float:right;
+  }
+}
+
+table#key {
+  td.space {
+    padding-left: 436px;
+  }
+  td.yellow {
+    font-size: 13px;
+    background: #fcf2c0;
+    padding: 5px 5px 5px 5px;
+  }
+  td.red {
+    font-size: 13px;
+    background: #ee9b9b;
+    padding: 5px 5px 5px 5px;
+  }
+  td.blue {
+    font-size: 13px;
+    background: #90C3D4;
+    padding: 5px 5px 5px 5px;
+  }
+  td.grey {
+    font-size: 13px;
+    background: #cfc9c9;
+    padding: 5px 5px 5px 5px;
   }
 }
 
--- a/app/assets/stylesheets/passengers.scss
+++ b/app/assets/stylesheets/passengers.scss
@@ -59,7 +59,7 @@ a {
 
 table#key {
   td.space {
-    padding-left: 436px;
+    padding-left: 339px;
   }
   td.yellow {
     font-size: 13px;
--- a/app/assets/stylesheets/passengers.scss
+++ b/app/assets/stylesheets/passengers.scss
@@ -11,7 +11,7 @@ body {
   line-height: 18px;
 }
 
-ol, ul, td {
+p, ol, ul, td {
   font-family: verdana, arial, helvetica, sans-serif;
   font-size: 13px;
   line-height: 18px;
@@ -23,7 +23,7 @@ strong.bold_and_underline {
   background-color: #e0dbfa;
   text-decoration: underline;
 }
-p {
+p.make_green {
   font-family: verdana, arial, helvetica, sans-serif;
   font-size: 19px;
   line-height: 18px;
@@ -56,26 +56,31 @@ a {
    float:right;
   }
 }
+table#key, table#passengers {
+  .will_expire_soon {
+    background-color: #fcf2c0;
+  }
+  .expired_within_grace_period {
+    background-color: #ee9b9b;
+  }
+  .no_note {
+    background-color: #cfc9c9;
+  }
+  .expired {
+    background-color: #90C3D4
+  }
+}
+.checkbox-inline.key_position {
+  float: right;
+}
 
 table#key {
-  td.space {
-    padding-left: 352px;
-  }
+  //td.space {
+  //  padding-left: 320px;
+  //}
   td {
     font-size: 13px;
     padding: 5px;
-  }
-  .expire_within_week {
-    background: #fcf2c0;
-  }
-  .expired_within_grace {
-    background: #ee9b9b;
-  }
-  .expired {
-    background: #90C3D4;
-  }
-  .no_note {
-    background: #cfc9c9;
   }
 }
 
@@ -85,9 +90,7 @@ table#passengers{
     text-align: center;
     padding: 5px 25px 5px 25px;
   }
-  tr {
-    border-bottom: 2px solid black;
-  }
+
   th {
     font-size: 20px;
     background: #9385d5;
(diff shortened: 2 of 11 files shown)


app/assets/stylesheets/passengers.scss
```diff
@@ -34,9 +46,44 @@ a {
...
+   float:right;
+  }
 }
 
-table{
+table#key {
+  td.space {
+    padding-left: 339px;
+  }
+  td.yellow {
```
werebus: I would suggest using the same class names in your key as the main table itself (`.expired_within_grace_period`, `.no_note`, etc.)  Also, since all your `table#key td`s have the same `font-size` and `padding`, you don't have to repeat that part. Make a less-specific rule:

```scss
table #key {
  td {
    font-size: 13px;
    padding: 5px;
  }
}
```

Generally, your classes should describe what kind of thing it is, not how they look.  Consider the following: what if ST asks you to make expiring-soon passengers salmon?  Then you're left with

```scss
table#key {
  td.yellow {
    background: LightSalmon; //which isn't yellow
  }
}
```

app/models/passenger.rb
```diff
@@ -7,11 +7,7 @@ class Passenger < ApplicationRecord
   scope :permanent, -> { where(permanent: true) }
   scope :temporary, -> { where(permanent: false) }
   scope :active, -> {where(active: true)}
-  scope :expired, -> {where(expired: false)}
-
-  def self.search(search)
```
werebus: Did this get moved? Or do you not need it because of DataTables?

app/models/passenger.rb
```diff
@@ -22,9 +18,15 @@ def self.deactivate_expired_doc_note
...
     expiration.present? && expiration <= 7.days.since && expiration >= Date.today
   end
-  def expired_within_3_days?
-    expiration.present? && expiration >= 4.days.ago && expiration <= 1.days.ago
+
+  def expired_within_grace_period? #will_expire_within_3_days?
+    expiration.present? && (expiration < Date.today && expiration >= 3.days.ago)
   end
+
+  def expired_for_3_days?
```
werebus: More than 3 days, perhaps?

app/views/passengers/index.haml
```diff
@@ -1,30 +1,54 @@
+%p
+  %b.yay
```
werebus: Two things:

1.  What does the class `yay` represent? What are we celebrating? I'm being a little sarcastic, but in truth this class name doesn't seem to be meaningful.
2.  Don't use `<b>` tags. ([generally](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong#Bold_vs._Strong)) Use `<strong>` if you actually want to emphasize the text over its surrounding text.  Otherwise, you can modify the `font-weight` on your class's css.

db/seeds.rb
```diff
@@ -26,27 +26,35 @@
                     wheelchair: false,
                     active: true,
                     permanent: true
+  Passenger.create! name: FFaker::Name.name,
```
werebus: Suggestion: You can use Ruby's iteration to DRY this up a bit

```ruby
[nil, Date.today + 2.days, Date.today - 2.days, Date.today - 4.days].each do |exp|
  Passenger.create! name: FFaker::Name.name,
                    address: FFaker::Address.street_address,
                    email: FFaker::Internet.email,
                    phone: FFaker::PhoneNumber.short_phone_number,
                    wheelchair: false,
                    active: true,
                    permanent: false,
                    expiration: exp
end
```

app/models/passenger.rb
Chauce: IIRC, I changed it to `scope :expired, -> {where(active: false)}`, because `active` being false was relevant to one of the filter elements.

werebus: Sorry, I meant the `search` method.

mboneil10:
Wow, this has changed a lot since I last saw it! I like it.

app/models/passenger.rb
```diff
@@ -22,9 +18,15 @@ def self.deactivate_expired_doc_note
   def will_expire_within_a_week?
     expiration.present? && expiration <= 7.days.since && expiration >= Date.today
   end
-  def expired_within_3_days?
-    expiration.present? && expiration >= 4.days.ago && expiration <= 1.days.ago
+
+  def expired_within_grace_period? #will_expire_within_3_days?
```
mboneil10: I see that you wrote a comment. I'm not sure what it means, but I'm guessing you're unsure of the method name.

Personally, I'd rather see `expired_within_grace_period?` than `will_expire_within_3_days?`. It helps me understand why you're checking if the thing (whatever it is) will expire within three days.

app/models/passenger.rb
mboneil10: Isn't it written as less than three days?

mboneil10: Isn't it written as less than three days?

mboneil10:
I like how the radio buttons are close to their corresponding text.

mboneil10:
It's hard for me to see the top left corner text. As in, it's a little bright. But I'm thinking that could be my screen? Not sure.

mboneil10:
Great job for putting this all together yourselves!

app/helpers/passengers_helper.rb
```diff
@@ -2,8 +2,10 @@ module PassengersHelper
   def expiration_check(passenger)
     if passenger.will_expire_within_a_week?
       'will_expire_soon'
-    elsif passenger.expired_within_3_days?
-      'expired_3_days'
+    elsif passenger.expired_within_grace_period?
+      'expired_within_grace_period'
+    elsif passenger.expired_for_3_days?
+      'expired_for_3_days'
```
Anbranin: Actually, @mboneil10 has a good point. If someday ST comes to us and says "hey, we want the grace period to be a week" how easy will it be to change? I'm thinking this could be a model constant somewhere. That way we'd only have to change it once.

app/controllers/passengers_controller.rb
```diff
@@ -9,11 +9,14 @@ def index
...
     @temporary = params[:filter] == 'temporary'
     @all = params[:filter] == 'all'
-    @search_term = params[:search]
+    @expired = params[:filter] == 'expired'
+    @active = params[:filter] == 'active'
     @passengers = Passenger.order :name
     @passengers = @passengers.search(@search_term) if @search_term.present?
     @passengers = @passengers.permanent if @permanent.present?
     @passengers = @passengers.temporary if @temporary.present?
+    @passengers = @passengers.expired if @expired.present?
```
dfaulken: Here too, you coerced to be true or false, so using `present?` is misleading.

app/helpers/passengers_helper.rb
werebus: While you're doing that, the 7 days for the warning would also probably be a good candidate for semi-configurability.

app/models/passenger.rb
avacmeyer: we don't need it anymore

app/helpers/passengers_helper.rb
avacmeyer: its super easy to change

app/models/passenger.rb
avacmeyer: it was the old name

app/assets/javascripts/datatables.js
```diff
@@ -0,0 +1,15 @@
...
+    { "bSortable": false },
+    { "bSortable": false },
+    null,
+    { "bSortable": false },
+    { "bSortable": false },
+    { "bSortable": false },
+  ],
+  "aLengthMenu": [[25, 50, 75, 100, -1], [25, 50, 75, 100, "All"]],
+  "iDisplayLength": 100
+});});
```
Chauce: `aLengthMenu` determines what options are available for displaying X number of passengers in the table. `iDisplayLength` makes it default to whatever number you put there, in this case, 100. We did that because it resets to whatever the default is when you filter things ... setting to 100 makes it less visually jarring.

werebus: Man, Datatables, what's up with your naming conventions? 😋

app/helpers/passengers_helper.rb
werebus: Do you understand @Anbranin's model-constant suggestion?

app/models/passenger.rb
mboneil10: Oh, ok. Good.

app/assets/javascripts/datatables.js
Chauce: I do not know but man they are weird!

app/views/passengers/index.haml
```diff
@@ -1,30 +1,54 @@
...
+          %td.expired_within_grace_period Expired Within Grace Period
+          %td.expired Expired
+          %td.no_note No Note
 %p
-= link_to "Add New Passenger", new_passenger_path
-
 - if @permanent
-  %p
-    %b Showing all Permanent
+  %p.make_green
```
Anbranin: What if they want you to make this a different color in the future? Then "make green" won't make any sense.

app/controllers/passengers_controller.rb
```diff
@@ -6,10 +6,10 @@ def new
   end
 
   def index
-    @permanent = params[:filter] == ‘permanent’
```
Anbranin: FUN FACT these weren't quotes

dfaulken:
I hadn't seen it before but @mboneil10 makes a great point. We should check all of the color combinations involved here for contrast ratios as defined by [WCAG 2.0 AA](https://www.w3.org/WAI/WCAG20/quickref/#visual-audio-contrast-contrast).

app/models/passenger.rb
```diff
@@ -7,24 +7,33 @@ class Passenger < ApplicationRecord
...
   end
 
-  def will_expire_within_a_week?
-    expiration.present? && expiration <= 7.days.since && expiration >= Date.today
+  def will_expire_within_warning_period?
+    expiration.present? && expiration < Passenger.expiration_warning && expiration >= Date.today
+  end
+
+  def expired_within_grace_period?
+    expiration.present? && (expiration < Date.today && expiration >= Passenger.grace_period)
```
dfaulken: Just FYI, these parentheses aren't needed.

`a & (b & c) = (a & b) & c = a & b & c`.

db/seeds.rb
```diff
@@ -10,43 +10,15 @@
...
-                    permanent: false,
-                    expiration: expirations.sample
-  Passenger.create! name: FFaker::Name.name,
-                    address: FFaker::Address.street_address,
-                    email: FFaker::Internet.email,
-                    phone: FFaker::PhoneNumber.short_phone_number,
-                    wheelchair: false,
-                    active: true,
-                    permanent: true
+20.times do
```
dfaulken: So you want 100 passengers whose registrations, in equal frequency:

1. Don't have an expiration
2. Expire in 2 days
3. Expired 2 days ago
4. Expired 4 days ago
5. Might've expired anytime in the last week, or might expire anytime in the next 6 months.

Am I reading that correctly?

app/controllers/passengers_controller.rb
werebus: Yup. Those are `&lsquo;` and `&rsquo;` ([U+2018][left] and [U+2019][right]). @jforgue ran into this a week or so ago when he was trying to set his MySQL password. Commonly, this comes from copy-and-pasting from someone's blog. Lots of blogging software transforms `'` into `‘` and `’` for *prose* because they look nicer, but code bloggers may be unaware of that fact.

[left]:  http://www.fileformat.info/info/unicode/char/2018/index.htm
[right]: http://www.fileformat.info/info/unicode/char/2019/index.htm

dfaulken: Yet another reason why copying and pasting code from the internet is counterproductive

app/assets/stylesheets/passengers.scss
```diff
@@ -16,7 +16,19 @@ p, ol, ul, td {
   font-size: 13px;
   line-height: 18px;
 }
-
+strong.bold_and_underline {
```
mboneil10: Maybe think of renaming `.bold_and_underline` to something else. For the same reason as @Anbranin's comment below

db/seeds.rb
Chauce: Not positive about being in equal frequency but yes, we're trying to cover all cases for expiration dates (at least, all cases for which we have color coding options).

dfaulken: Okay. I ask because covering all your bases is good for seed data, but so is resembling production data. It seemed like a weird distribution of users that wouldn't actually exist. In particular, having the vast majority of your users having extremely proximate expiration dates doesn't seem quite right.

dfaulken: That being said, that concern is not a functional one, purely a question of preference.

PR #5, opened 2017-06-28
Admin

Chauce:
screen shot 2017-06-28 at 10 49 41 am
New development login
screen shot 2017-06-28 at 10 50 13 am
Admin view shows 'delete' option
screen shot 2017-06-28 at 10 50 31 am
Regular dispatcher view does not show delete option
screen shot 2017-06-28 at 10 51 55 am
Adding permanent passengers disabled for regular dispatchers. Regular dispatchers also cannot see the 'active' field (this is taken care of by a chron job that checks the doctor's note expiration date, but admins can override it to reactivate a passenger)
screen shot 2017-06-28 at 10 55 35 am
Admins can add permanent passengers and see the 'active' checkbox

PR #7, opened 2017-07-01
Fix admin

dfaulken:
Ref: https://github.com/umts/st-pax/pull/5/commits/9bd1f17ac19d14fb170c5bf941e0d888566e5939

Also took care of a few things I ran across while getting an app instance up and running.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,7 @@ group :development do
   # Access an IRB console on exception pages or by using <%= console %> anywhere in the code.
   gem 'web-console', '>= 3.3.0'
   gem 'listen', '>= 3.0.5', '< 3.2'
+  gem 'rb-fsevent', '0.9.8'
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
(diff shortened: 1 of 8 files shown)


Gemfile
```diff
@@ -48,6 +48,7 @@ group :development do
   # Access an IRB console on exception pages or by using <%= console %> anywhere in the code.
   gem 'web-console', '>= 3.3.0'
   gem 'listen', '>= 3.0.5', '< 3.2'
+  gem 'rb-fsevent', '0.9.8'
```
dfaulken: https://github.com/guard/listen/issues/431

werebus: (Which appears to be fixed in rb-fsevent 0.10.2)

Chauce: Ah, I was wondering what was up with that!

Anbranin: Why don't we update rb-fsevent then?

dfaulken: Sure, go nuts.

dfaulken:
Cool. Merge at your leisure.

PR #10, opened 2017-07-25
Mobility

Chauce:
screen shot 2017-07-25 at 2 15 31 pm
screen shot 2017-07-25 at 2 16 59 pm

Added a dropdown option of mobility devices as per ST request.

### DIFF
--- a/app/models/passenger.rb
+++ b/app/models/passenger.rb
@@ -10,7 +10,8 @@ class Passenger < ApplicationRecord
   scope :inactive, -> { where(active: false) }
 
   # TODO: Make configurable by user
-  MOBILITY_DEVICES = ['Boot', 'Crutches', 'Cane', 'Walker', 'Service Dog']
+  MOBILITY_DEVICES = ['Boot', 'Crutches', 'Cane', 'Walker',
+                      'Service Dog'].freeze
 
   before_save do
     assign_attributes(expiration: nil) if permanent?
--- a/app/models/passenger.rb
+++ b/app/models/passenger.rb
@@ -9,6 +9,9 @@ class Passenger < ApplicationRecord
   scope :active, -> { where(active: true) }
   scope :inactive, -> { where(active: false) }
 
+  # TODO: Make configurable by user
+  MOBILITY_DEVICES = ['Boot', 'Crutches', 'Cane', 'Walker', 'Service Dog']
+
   before_save do
     assign_attributes(expiration: nil) if permanent?
   end
--- a/app/views/passengers/_form.html.haml
+++ b/app/views/passengers/_form.html.haml
@@ -22,8 +22,9 @@
     = form.label :wheelchair
     = form.check_box :wheelchair
   .field
-    = form.label :mobility_devices
-    = form.select :mobility_devices
+    = form.label :mobility_device
+    = form.select :mobility_device, Passenger::MOBILITY_DEVICES,
+    include_blank: 'None'
   - if @current_user.admin?
     .field
       = form.label :active
--- a/app/views/passengers/_form.html.haml
+++ b/app/views/passengers/_form.html.haml
@@ -21,6 +21,9 @@
   .field
     = form.label :wheelchair
     = form.check_box :wheelchair
+  .field
+    = form.label :mobility_devices
+    = form.select :mobility_devices
   - if @current_user.admin?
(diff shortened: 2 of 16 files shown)


app/views/passengers/index.haml
```diff
@@ -48,7 +48,7 @@
     %b Showing All Passengers
 %p
 - if @passengers.present?
-  %table#passengers{class: table_class}
+  %table#passengers{class: passengers_table_class}
```
werebus: Probably don't need `_class` on your class name.

app/views/passengers/show.haml
```diff
@@ -17,6 +17,12 @@
   - else
     No
 %p
+  %b Mobility Device:
+  - if @passenger.mobility_device?
```
werebus: Does this work?  Don't you mean `@passenger.mobility_device.present?`

app/views/passengers/index.haml
dfaulken: It's a helper method call

werebus: Yup.  Disregard.

app/models/passenger.rb
```diff
@@ -9,6 +9,10 @@ class Passenger < ApplicationRecord
   scope :active, -> { where(active: true) }
   scope :inactive, -> { where(active: false) }
 
+  # TODO: Make configurable by user
```
dfaulken: The way do to that, when you get there, is to write a `MobilityDevice` model.

db/seeds.rb
```diff
@@ -27,7 +28,7 @@
   User.create! name: FFaker::Name.name,
                email: FFaker::Internet.email,
                phone: FFaker::PhoneNumber.short_phone_number,
-               spire: rand(10 ** 8).to_s.rjust(8, '0'),
+               spire: rand(10**8).to_s.rjust(8, '0'),
```
dfaulken: Oh really? Did Rubocop not like this one?

werebus: Yup:  https://github.com/bbatsov/ruby-style-guide#spaces-operators

>  There are a few exceptions. One is the exponent operator ... Another exception is the slash in rational literals,

app/models/passenger.rb
```diff
@@ -9,6 +9,10 @@ class Passenger < ApplicationRecord
...
+  MOBILITY_DEVICES = ['Boot', 'Crutches', 'Cane', 'Walker',
```
mboneil10: I was asking @Anbranin this question earlier... doesn't a wheelchair count as a mobility device?

app/views/passengers/_form.html.haml
```diff
@@ -21,6 +21,10 @@
   .field
     = form.label :wheelchair
     = form.check_box :wheelchair
+  .field
+    = form.label :mobility_device
+    = form.select :mobility_device, Passenger::MOBILITY_DEVICES,
```
mboneil10: Nice job on select

db/seeds.rb
dfaulken: Alrighty then Bozhidar

app/views/passengers/show.haml
Chauce: It does actually work ... but I'll change it to the proper syntax.

werebus: Huh.  I don't know if one is more "correct" than the other.  I honestly had no idea that there were predicate methods defined for *all* attributes.

app/models/passenger.rb
Chauce: Making a note so that I remember to check in with ST about this because I don't 100% remember what the logic was on this one.

mboneil10: This is related to @dfaulken's comment about configurability. 👍

app/models/passenger.rb
Chauce: Sounds good! We're going to tackle this in a different branch so there isn't too much going on in one branch at one time.

app/models/passenger.rb
Chauce: This will be getting sorted in a different branch - just going to leave it as a model constant for this branch.

werebus:
This ended up having the `MobilityDevice` model in it too, but it's not doing any harm  in there waiting to be implemented.

PR #12, opened 2017-08-15
Docnote form

Chauce:
Changed from having "expiration date" be a column in Passengers table, to now being a column in the new DoctorsNote table. Had to make various changes to reflect the new functionality. You can now input doctor's notes expiration dates for passengers (new and as updates), just like you could before! Lots of new stuff under the hood but the outside looks ... pretty much the same. Like putting a V8 engine in a Honda.


Gemfile
```diff
@@ -54,6 +54,3 @@ group :development do
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
 end
-
-# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
-gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
```
dfaulken: Okay. There's no harm for our purposes in removing this, but there's also no harm leaving it there.

app/controllers/passengers_controller.rb
```diff
@@ -3,6 +3,15 @@ class PassengersController < ApplicationController
   before_action :access_control, only: %i[destroy]
   def new
     @passenger = Passenger.new
+    @doctors_note = DoctorsNote.new
+  end
+
+  def edit
+    if @passenger.doctors_note.present?
```
dfaulken: Could also be

```ruby
@doctors_note = @passenger.doctors_note || DoctorsNote.new
```

Is it an issue that `DoctorsNote.new` is not associated with the passenger, whereas `@passenger.doctors_note` would be?

app/models/doctors_note.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def self.expiration_warning
+    7.days.since.to_date
+  end
+
+  def self.deactivate_expired_doc_note
+    active.where('expiration_date < ?', grace_period).each do |doctors_note|
+      doctors_note.update_attributes active: false
+    end
+  end
```
dfaulken: It doesn't look like doctors notes have an `active` column.

app/models/doctors_note.rb
```diff
@@ -0,0 +1,41 @@
...
+      expiration_date >= DoctorsNote.grace_period
+  end
+
+  def expired?
+    expiration_date.present? && expiration_date < DoctorsNote.grace_period
+  end
+
+  def temporary?
+    !passenger.permanent?
+  end
```
dfaulken: This method doesn't seem like it belongs on this class.

spec/factories/doctors_notes.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :doctors_note do
+    passenger_id 1
+    expiration_date "2017-07-27"
+    overriden_by 1
```
dfaulken: Curious what this column represents. If it's a user, then I'd recommend defining an association that way in your model file.

Gemfile
Anbranin: There is, actually, harm in leaving it there. @AkivaGreen had a problem installing this gem so @werebus and I just told him to remove it. We could either spend time troubleshooting why a gem used solely for windows is malfunctioning, or just take it out.

spec/factories/doctors_notes.rb
AkivaGreen: Well, I was thinking of just leaving the factory with one line for the passenger association since that is the only one that would be needed. Therefore, everything else would be filled out in the tests as needed.

Gemfile
dfaulken: Okay. I thought `platforms` meant that it would only run if one was using those platforms.

spec/factories/users.rb
```diff
@@ -1,7 +1,7 @@
 FactoryGirl.define do 
   factory :user do 
-    sequence(:name) {|n| "Name #{n}"}
-    sequence(:email) {|n| "test#{n}@example.com"}
+    name FFaker::Name.name
```
dfaulken: Doesn't this mean that every factory-generated user will have the same fake name?

spec/factories/doctors_notes.rb
dfaulken: Right, I said in your *model*. I see no association matching this there.

Gemfile
werebus: Hmm... You know, you're right.

To be fair, I'm not sure that @AkivaGreen's problem was "installing this gem.'"  I just head him say "tzinfo" and I said, "Isn't that Windows-only? Chuck it!"

app/controllers/passengers_controller.rb
dfaulken: Great!

spec/factories/users.rb
dfaulken: That is, it should be `name { FFaker::Name.name }` (in a block)

app/views/passengers/index.haml
```diff
@@ -64,7 +65,7 @@
           %th
     %tbody
       - @passengers.each do |passenger|
-        %tr{class: expiration_check(passenger)}
+        %tr{class: passenger.doctors_note.present? ? expiration_check(passenger.doctors_note) : ""}
```
AkivaGreen: I think it would be best to just call the method and check for nil in the method itself so we can handle it better.

Gemfile
AkivaGreen: I was getting errors in my bundler and it was because I need to upgrade my ruby version. After upgrading the version this solved itself.

dfaulken:
@AkivaGreen and @Anbranin, do tests need to be part of this PR? It's already pretty big.

PR #16, opened 2017-08-17
Dispatch log

dfaulken:
Implement dispatch log functionality. Dispatchers can create log entries and delete their own entries should they make a mistake. Admins can delete and edit any entry, and create entries themselves if they so wish.

It doesn't look that nice right now:

screen shot 2017-08-17 at 2 10 45 pm

So perhaps a next good step would be to apply some styling, but before that I thought I'd run the basic functionality past everyone.

Master seems broken right now, so I built off of `docnote-form`, which will likely be merged before this anyway.


Gemfile
```diff
@@ -40,6 +40,7 @@ group :development, :test do
   gem 'ffaker'
   gem 'pry-byebug'
   gem 'rubocop'
+  gem 'timecop'
```
dfaulken: Added for seeding purposes, so that log entries could date back to a month or so ago.

app/assets/stylesheets/log.scss
```diff
@@ -0,0 +1,7 @@
+ul.dispatch-log{
+  li{
+    .edit-form{
+      display: none;
+    }
+  }
+}
```
dfaulken: Could've written as `ul.dispatch-log li .edit-form`, but I anticipate more styling will be added to this file.

app/models/user.rb
```diff
@@ -1,6 +1,12 @@
 class User < ApplicationRecord
   validates :name, :email, :spire, presence: true
 
+  scope :dispatchers, -> { where.not admin: true }
+
+  def can_delete?(item)
+    admin? || (item.is_a?(LogEntry) && item.user == self)
+  end
```
dfaulken: I could also move this to the `LogEntry` class as `can_be_deleted_by?`.

config/application.rb
```diff
@@ -28,5 +28,7 @@ class Application < Rails::Application
 
     # Don't generate system test files.
     config.generators.system_tests = nil
+
+    config.time_zone = 'Eastern Time (US & Canada)'
```
dfaulken: Unrelated to the content of my PR, but should've been here to begin with.

db/seeds.rb
```diff
@@ -1,13 +1,21 @@
...
+20.times { create :passenger, :permanent }
+20.times { create :passenger, :recently_expired }
+20.times { create :passenger, :expiring_soon }
+20.times { create :passenger, :expiration_overriden }
+20.times { create :passenger, :inactive }
 5.times  { create :passenger, :no_note }
+
+dispatchers = User.dispatchers
+30.times do
+  Timecop.freeze 1.month.ago + rand(1.month) do
```
dfaulken: Should I include a comment explaining what this does? Or is it clear?

app/models/user.rb
```diff
@@ -1,6 +1,12 @@
 class User < ApplicationRecord
   validates :name, :email, :spire, presence: true
 
+  scope :dispatchers, -> { where.not admin: true }
```
dfaulken: Used in seeds, but potentially useful elsewhere.

db/seeds.rb
jforgue: I think a comment would be good here, even if it is pretty clear it'll help us new programmers avoid confusion down the road.

dfaulken:
Could you highlight a few places that you think could use comments?

app/models/user.rb
Anbranin: Nah. I mean it's a little weird but the other ways of doing this don't make more sense.

Issue #6, opened 2017-06-29
Test application


Anbranin:
This application needs some tests. Also SimpleCov to track coverage. And let's set up Rubocop as well.

mboneil10:
We're waiting for system tests to be in Rspec.

PR #21, opened 2017-08-17
Tests

AkivaGreen:
Some tests for #6


spec/features/passenger_management_spec.rb
```diff
@@ -1,12 +1,29 @@
 require 'rails_helper' 
 
 feature 'Passenger Management' do 
+  before :each do
+    sign_in(create :user, :admin)
+  end
   scenario 'creating a new passenger' do 
+    visit passengers_path
+    click_link 'Add New Passenger'
+    fill_in('passenger[name]', with: 'Foo Bar')
```
dfaulken: Normally for these if the form is constructed correctly you can say `fill_in 'Name', with: 'Foo Bar'`.

spec/models/passenger_spec.rb
```diff
@@ -38,4 +38,14 @@
       expect(@passenger.temporary?).to eql true
     end
   end
+
+  describe 'self.deactivate_expired_doc_note' do
```
dfaulken: This method depends on the grace period, so if that's changed this test could be affected. It'd be nice for the fact that this method uses an external method call as a data source to be reflected in its test.

spec/features/passenger_management_spec.rb
AkivaGreen: It will give me an error saying it can't find the fields if I did 'Name' or 'Passenger name' or even just 'name'.

dfaulken: Right, that's why I led with "if the form is constructed correctly". What's here is fine, though.

app/controllers/passengers_controller.rb
```diff
@@ -44,7 +44,7 @@ def update
 
   def destroy
     @passenger.destroy
-    redirect_to passengers_url, notice: 'Passenger was successfully destroyed.'
+    redirect_to passengers_path, notice: 'Passenger was successfully destroyed.'
```
dfaulken: Why this change?

spec/features/passenger_management_spec.rb
```diff
@@ -2,13 +2,58 @@
 
 require 'rails_helper'
 
-feature 'Passenger Management' do
-  scenario 'creating a new passenger' do
+feature 'Passenger Management' do 
+  feature 'as and admin' do
```
dfaulken: Should be `as an admin`

spec/features/passenger_management_spec.rb
```diff
@@ -2,13 +2,58 @@
...
 require 'rails_helper'
 
-feature 'Passenger Management' do
-  scenario 'creating a new passenger' do
+feature 'Passenger Management' do 
+  feature 'as and admin' do 
+    before :each do
+      sign_in(create :user, :admin)
+    end
+    scenario 'creating a new passenger' do
```
dfaulken: Will this `feature` and `scenario` DSL translate well to system tests? It should be fine to just stick to `describe` and `it`, particularly since that gives verbal descriptions of the expected behavior.

Will defer to @Anbranin here.

spec/models/log_entry_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper' 
+
+describe LogEntry do 
+  describe 'entry_time' do 
+    before do 
+      Timecop.freeze(Time.local(2017, 03, 16, 10, 30, 0))
```
dfaulken: Since we only have one test here, it might be nicer to use the `Timecop.freeze ... do` block syntax, so that we don't need `before` and `after` blocks.

spec/models/passenger_spec.rb
dfaulken: To clarify this comment, we should use a mock here.

spec/models/user_spec.rb
```diff
@@ -6,9 +6,44 @@
...
     it 'gets the not of admin' do
       expect(@user.dispatcher?).to eql false
     end
   end
+
+  describe 'can_delete?' do 
+    context 'when admin' do 
+      it 'returns true' do 
+        item = create :log_entry
+        expect(@user.can_delete?(item)).to eql true
```
dfaulken: For `true` and `false` (and a few other things) you can write `to be` rather than `to eql`. Source: https://github.com/umts/custom-cops/blob/master/lib/umts-custom-cops/be_matcher_for_non_duplicable_types.rb

spec/spec_helper.rb
```diff
@@ -1,3 +1,23 @@
+# This file was generated by the `rails generate rspec:install` command. Conventionally, all
```
dfaulken: We can remove these comments.

spec/spec_helper.rb
```diff
@@ -11,6 +31,6 @@
   config.shared_context_metadata_behavior = :apply_to_host_groups
 end
 
-def sign_in(user)
-  session[:user_id] = user.id
+def sign_in(user) 
+  page.set_rack_session(user_id: user.id)
```
dfaulken: To remove some duplication, I typically like to make this method such that you can just pass it a symbol (perhaps a factory trait) or a User instance, and it will take care of creating a user for you if necessary, Cf. https://github.com/umts/umasstransit.rodeo/blob/6feb613407a9a04caa434665f2eab525184e57c6/spec/spec_helper.rb#L23-L35

dfaulken:
Most of my comments are 'outdated' because the lines were changed afterwards, but they still apply in general;.

spec/models/log_entry_spec.rb
```diff
@@ -4,16 +4,12 @@
...
-      Timecop.freeze(Time.local(2017, 0o3, 16, 10, 30, 0))
-    end
-    after do
-      Timecop.return
-    end
     it 'returns a string of created_at date and time' do
       time = '10:30 am, Thursday, March 16, 2017'
-      log_entry = create :log_entry
-      expect(log_entry.entry_time).to eql time
+      Timecop.freeze(Time.local(2017, 0o3, 16, 10, 30, 0)) do
```
dfaulken: The fix here is to just write `3`

app/views/passengers/_form.html.haml
```diff
@@ -19,11 +19,8 @@
     = form.label :phone_number
     = form.phone_field :phone
   .field
-    = form.label :wheelchair
-    = form.check_box :wheelchair
-  .field
-    = form.label :mobility_device
-    = form.select :mobility_device, Passenger::MOBILITY_DEVICES,
+    = form.label :mobility_device_id
+    = form.select :mobility_device_id, MobilityDevice.all.collect { |md| [md.device, md.id] },
```
dfaulken: Should use `options_from_collection_for_select` here

app/views/passengers/show.haml
```diff
@@ -11,15 +11,9 @@
...
-  - if @passenger.wheelchair
-    Yes
-  - else
-    No
-%p
   %b Mobility Device:
-  - if @passenger.mobility_device?
-    = @passenger.mobility_device
+  - if @passenger.mobility_device.present?
+    = @passenger.mobility_device.device
```
dfaulken: `:try` will work here also

spec/models/log_entry_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe LogEntry do
+  describe 'entry_time' do
+    it 'returns a string of created_at date and time' do
+      time = '10:30 am, Thursday, March 16, 2017'
+      Timecop.freeze(Time.local(2017, 0o3, 16, 10, 30, 0)) do
```
dfaulken: The fix here is to just write `3`, rather than an octal number

dfaulken: Also you don't need to specify `0` seconds, if you just leave the seconds out it'll be zero by default

spec/models/passenger_spec.rb
```diff
@@ -36,7 +36,18 @@
...
     it 'gets the not of permanent' do
-      expect(@passenger.temporary?).to eql true
+      expect(@passenger.temporary?).to be true
+    end
+  end
+
+  describe 'self.deactivate_expired_doc_note' do
+    it 'deactivates the expired passenger' do
+      create :doctors_note, passenger: @passenger,
+                            expiration_date: DoctorsNote.grace_period - 1.day
```
dfaulken: Should use a mock here rather than actually calling a method this code depends on

spec/spec_helper.rb
```diff
@@ -11,6 +17,17 @@
...
-  session[:user_id] = user.id
+def when_current_user_is(user)
+  current_user = 
+    case user
+    when Symbol then create :user, user 
+    when User then user 
+    when nil then nil
+    else raise ArgumentError, 'Invalid user type'
+    end
+    login_as current_user
```
dfaulken: Should be indented one level fewer

spec/models/log_entry_spec.rb
werebus: Yeah, I wish Rubocop would be more specific that you only want to put the `o` in there if you *really want* an octal number.  Hint: no one ever really wants an octal number.

AkivaGreen:
Hey @dfaulken, this is all good to look at again. However, there is one test failing and that is the inactive filter feature spec. The current implementation of the filters doesn't work for showing only inactive passengers. Should I just leave it be or should I just fix it so I can have the test pass?

dfaulken:
Up to @Anbranin, but I think the current functionality is fine. Personally I would just remove the test.

app/views/passengers/show.haml
```diff
@@ -11,15 +9,9 @@
...
-  - if @passenger.wheelchair
-    Yes
-  - else
-    No
-%p
   %b Mobility Device:
-  - if @passenger.mobility_device?
-    = @passenger.mobility_device
+  - if @passenger.try(:mobility_device)
+    = @passenger.mobility_device.device
```
dfaulken: Just `passenger.mobility_device.try(:device)`, with no conditional, will work.

If there is a mobility device, then `try` calls `#device` on it. If there is not (it is `nil`) then it just returns `nil`.

coverage/.last_run.json
```diff
@@ -0,0 +1,5 @@
+{
+  "result": {
+    "covered_percent": 96.52
```
dfaulken: Where are we missing coverage?

spec/features/passenger_filters_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+feature 'Passenger Filters' do
```
dfaulken: In general these tests don't ensure that the correct passenger is shown, just that the correct number is shown. It'd be nice to be sure that exactly the right data is present.

spec/models/passenger_spec.rb
```diff
@@ -9,34 +9,34 @@
...
-      it "returns 'No Note'" do
-        expect(@passenger.expiration_display).to eql 'No Note'
-      end
-    end
   end
 
   describe 'temporary?' do
     it 'gets the not of permanent' do
-      expect(@passenger.temporary?).to eql true
+      expect(@passenger.temporary?).to be true
```
dfaulken: `expect(passenger).to be_temporary`

spec/models/passenger_spec.rb
```diff
@@ -9,34 +9,34 @@
...
+  end
+
+  describe 'self.deactivate_expired_doc_note' do
+    it 'deactivates the expired passenger' do
+      create :doctors_note, passenger: @passenger,
+                            expiration_date: 4.days.ago
+      @passenger.update active: true
+      Passenger.deactivate_expired_doc_note
+      @passenger.reload
+      expect(@passenger.active).to be false
```
dfaulken: `expect(passenger).not_to be_active`

app/views/passengers/show.haml
AkivaGreen: It will just leave it blank on the page if there is no mobility device instead of showing 'None'. Is that fine?

AkivaGreen: Nvm, I remembered a way to work on that

`= @passenger.mobility_device.try(:device) || 'None'`

spec/models/passenger_spec.rb
AkivaGreen: So Rspec can take the boolean columns and test them in this way? That's pretty cool.

dfaulken: It's even more generic than that — you can write `be_` followed by anything, and it will call that something as a predicate method (so `be_bananas` calls `bananas?`).

PR #26, opened 2017-08-21
Remove email from users

### LINKED ISSUE #23: Why do users need emails?
I don't see them used anywhere.

Since users missing them are "invalid" from Access, may as well remove the attribute and validation if we don't need it for anything.

dfaulken:
Closes #23.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170801202951) do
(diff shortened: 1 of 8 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170801202951) do
+ActiveRecord::Schema.define(version: 20170821165344) do
```
dfaulken: Frickin' schema every frickin' time

Issue #19, opened 2017-08-17
Data migration


dfaulken:
Before deployment, we need to figure out how to import existing data from the MS Access database into this app. Ideally in production the initial procedure would be to:

1. Create the database.
2. Schema load.
3. Run the task which imports the data exported from Access.

@sherson says he or @werebus can get us a CSV to develop an import with, and then another at time of initial deployment.

We also need to talk to ST about whether they wish to perform dual-entry of passengers or migrate instantaneously from Access to this interface.

dfaulken:
Will unassign @sherson once we have a test CSV; will unassign @Anbranin once we've had that conversation with ST.

dfaulken:
Thanks @sherson! Got a bunch of CSVs out of MS Access today: https://umass.app.box.com/files/0/f/35571889379/st-pax_production_CSVs

I'll start working on importing what we need now.

dfaulken:
After we import the data, we'll have two validation errors to fix.

1. Dispatchers (users) do not have SPIRE IDs in Access. We can either set them ourselves from round-three, or set just Andrew's / Avia's, and they can have the experience of filling in the rest.
1. "Diana Noble" and "Student Staff" are the only two admins, since the two STSPVs are in the database as just ST Dispatcher (well, one of them is ST Dspatcher). We will need to manually elevate them to be admins.

Any preference for how we handle SPIREs @Anbranin?

PR #28, opened 2017-08-22
CSV import, first steps

dfaulken:
Before I go too much further in this vein, I thought I'd get feedback on how it looks so far.

Intended procedure is to `rails db:schema:load` and `rails import:all` with [the necessary CSV files](https://umass.app.box.com/files/0/f/35571889379/st-pax_production_CSVs) in the data folder.

Addresses #19.

### DIFF
--- a/lib/tasks/import/dispatchers.rake
+++ b/lib/tasks/import/dispatchers.rake
@@ -15,11 +15,10 @@ namespace :import do
         puts "disp.csv:#{line} : Unrecognized status '#{row.fetch('Status')}'"
         next
       end
-      access_id = row.fetch('ID').to_i
       last_name, first_name = row.fetch('dispatcher').split ', '
       name = [first_name, last_name].join ' '
 
-      user = User.new name: name, admin: admin, access_id: access_id
+      user = User.new name: name, admin: admin
       unless user.save
         puts "disp.csv:#{line} : Skipping validations."
         user.save validate: false
--- a/lib/tasks/import/dispatchers.rake
+++ b/lib/tasks/import/dispatchers.rake
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+desc 'Import dispatchers from Access CSV at data/disp.csv'
+namespace :import do
+  task dispatchers: :environment do
+    CSV.foreach 'data/disp.csv', headers: true do |row|
+      binding.pry
+    end
+  end
+end
--- a/lib/tasks/import/dispatchers.rake
+++ b/lib/tasks/import/dispatchers.rake
@@ -5,8 +5,29 @@ require 'csv'
 desc 'Import dispatchers from Access CSV at data/disp.csv'
 namespace :import do
   task dispatchers: :environment do
-    CSV.foreach 'data/disp.csv', headers: true do |row|
-      binding.pry
+    CSV.foreach('data/disp.csv', headers: true).with_index 2 do |row, line|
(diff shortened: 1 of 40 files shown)


lib/tasks/import/dispatchers.rake
```diff
@@ -0,0 +1,32 @@
...
+require 'csv'
+
+desc 'Import dispatchers from Access CSV at data/disp.csv'
+namespace :import do
+  task dispatchers: :environment do
+    CSV.foreach('data/disp.csv', headers: true).with_index 2 do |row, line|
+      case row.fetch 'Status'
+      when 'ST Coordinator', 'Student Staff'
+        admin = true
+      when 'ST Dispatcher', 'ST Dspatcher', 'ST Dispatcer'
```
dfaulken: I kid you not

lib/tasks/import/dispatchers.rake
```diff
@@ -0,0 +1,32 @@
...
+      user = User.new name: name, admin: admin
+      unless user.save
+        puts "disp.csv:#{line} : Skipping validations."
+        user.save validate: false
+      end
+    end
+
+    puts "Created #{User.count} users."
+    puts "#{User.count(&:valid?)} are valid."
+    puts
```
Anbranin: just puts?

lib/tasks/import/log.rake
```diff
@@ -0,0 +1,24 @@
...
+
+desc 'Import daily log from Access CSV at data/daily_log.csv'
+namespace :import do
+  task log: :environment do
+    CSV.foreach('data/daily_log.csv', encoding: 'ISO-8859-1', headers: true).with_index 2 do |row, line|
+      date = Date.parse row.fetch('Date:')
+      user_name = row.fetch('Dispatcher:')
+      if user_name.is_a? String
+        last_name, first_name = user_name.split(', ')
+        user = User.find_by name: [first_name, last_name].join(' ')
```
Anbranin: The user that created the log won't always exist, unless we're importing Dispatchers that don't work here any more.

lib/tasks/import/dispatchers.rake
dfaulken: Newline-separates each task's output when running `import:all`

lib/tasks/import/log.rake
dfaulken: We can't import prior dispatchers - they're removed from the table (rather than set as inactive).

Your diagnosis is correct — these imports will result in a good deal of invalid data. Up to us to determine how to either correct the data or handle its invalidness.

Anbranin: I can see that it won't throw an error if user is nil, so that's okay.

I think it may be important to know who submitted prior log entries - would it make sense to just add a line that asks if user is nil, add the username like `" - submitted by #{user_name}" at the end of the note?

werebus: It *may* be important, but that's just too bad.

The dispatcher table only has current dispatchers in it. STSPVs have been deleting users as they leave transit. So, the "who" in those cases is unknowable. There's nothing but an orphaned foreign key that is only used in in dispatcher database.

dfaulken: Correct. More infuriatingly, existing dispatchers are referenced by name in the table — so if non-existing dispatchers were referenced by name too, we could do that. But they're just IDs, hence my check for `is_a? String`. So it's either a name, from which I have to do a DB lookup, or an ID that doesn't help me. 🎉

PR #32, opened 2017-08-23
Handle invalid data

### LINKED ISSUE #19: Data migration

dfaulken:
Closes #19.

Turns out, this app is fairly permissive about missing attributes.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -14,8 +14,8 @@ class SessionsController < ApplicationController
 
   def dev_login # route not defined in production
     if request.get?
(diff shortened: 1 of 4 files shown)


dfaulken:
Eh this is so trivial I'll just merge

PR #34, opened 2017-08-23
Refactor passenger filtering

### LINKED ISSUE #30: Hide inactive passengers by default
Particularly given that we will soon have data stretching back well over a decade.

dfaulken:
Closes #30.

screen shot 2017-08-23 at 2 02 05 pm

Default is show both permanent & temporary, and hide inactive (show only active). You can select whether to see temporary, permanent, or both; and then (separately) whether to show inactive.

Previously was just one radio button, which didn't allow for combining the filters.

### DIFF
--- a/app/views/passengers/index.haml
+++ b/app/views/passengers/index.haml
@@ -7,17 +7,14 @@
   %p
   = label_tag 'Filter by Type:'
   .passenger-filters
-    Permanent
-    = check_box_tag :filter, 'permanent', @permanent
-    .checkbox-inline
-    Temporary
-    = check_box_tag :filter, 'temporary', @temporary
-    .checkbox-inline
-    Inactive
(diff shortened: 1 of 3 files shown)


app/views/passengers/index.haml
```diff
@@ -6,20 +6,15 @@
...
-  .checkbox-inline
-  Show All
-  = radio_button_tag :filter, 'all'
+  .passenger-filters
+    Permanent Only
+    = radio_button_tag :filter, 'permanent', @permanent
+    Temporary Only
+    = radio_button_tag :filter, 'temporary', @temporary
+    All
+    = radio_button_tag :filter, '', !(@permanent || @temporary)
```
dfaulken: Open to making another instance variable out of this with an `else` in the first conditional in the controller.

PR #43, opened 2017-08-27
Replace recently

### LINKED ISSUE #42: s/Recently/Within Grace Period
The code and views use the word "recently" a lot to mean "within the grace period" for expiration dates of doctor's notes. This tripped me up during a refactor to the point where [we thought other things were broken](https://github.com/umts/st-pax/issues/40), so let's remove the ambiguity and replace the word "Recently" with "Within Grace Period" across the site (code included).

mboneil10:
Closes #42


mboneil10:
Not sure if the readme needed to be changed as well.

spec/factories/passengers.rb
```diff
@@ -18,9 +18,9 @@
       doctors_note nil
     end
 
-    trait :recently_expired do
+    trait :expired_within_grace_period do
```
dfaulken: The seed file *does* need to be changed. It now references a factory which no longer exists.

mboneil10: Ooops

PR #45, opened 2017-08-28
Capistrano

### LINKED ISSUE #20: Capistrano
Once we decide where we're going to deploy the app, someone needs to set up Capistrano to match.

mboneil10:
Closes #20 

I also did some rubocop stuff.


config/deploy/production.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+remote_user = Net::SSH::Config.for('af-transit-app3.admin.umass.edu')[:user] || ENV['USER']
```
dfaulken: This will change, but we can merge for now.

config/deploy.rb
```diff
@@ -0,0 +1,26 @@
...
+set :repo_url, 'git@github.com:umts/st-pax.git'
+set :branch, :master
+
+set :keep_releases, 5
+
+set :deploy_to, "/srv/#{fetch :application}"
+
+set :log_level, :info
+
+set :whenever_command, %i[sudo bundle exec whenever]
```
dfaulken: We should make sure that `PATH` is set correctly in the `config/schedule.rb` file.

See https://github.com/umts/jobapps/pull/300.

mboneil10: I don't. What.

mboneil10: Literally, just that line?

dfaulken: Yes, we should add that line.

dfaulken:
@mboneil10 I merged #51 to remove some of the noise from this PR.

PR #46, opened 2017-08-28
Add mobility device to passengers table

### LINKED ISSUE #44: Show mobility device in passengers table?
If we show the mobility device in the passengers table, then we show all of their attributes.

So perhaps we can remove the 'Show' link entirely, except for admins, who should be able to see the additional attributes in #27.

Seems like that's a thing dispatchers would want to just see without having to click through every time someone calls.

dfaulken:
Closes #44.

Also used the yes/no images, and only showed the 'Active' column if you were showing inactive people (otherwise why show a green checkmark for everyone in the table?).

screen shot 2017-08-28 at 3 58 20 pm


app/views/passengers/index.haml
```diff
@@ -43,16 +45,12 @@
       - @passengers.each do |passenger|
         %tr{class: passengers_table_row_class(passenger)}
           %td= passenger.name
+          %td= passenger.mobility_device.try(:device)
```
mboneil10: The passenger wouldn't have a `mobility_device`? Or is this for safety reasons?

db/seeds.rb
```diff
@@ -4,14 +4,14 @@
...
 
-%w[Cane Crutches Wheelchair].each do |device|
-  create :mobility_device, device: device
-end
-
 2.times { create :user, :admin }
 
 5.times { create :user }
 
+%w[Cane Crutches Wheelchair].each do |device|
```
mboneil10: Why are there only three? I thought there were four or five?

app/views/passengers/index.haml
```diff
@@ -30,9 +30,11 @@
     %thead
       %tr
         %th Name
+        %th Mobility Device
         %th Email
         %th Phone
-        %th.yesno Active
+        - unless @filters.include? 'Active'
```
mboneil10: Why is the filter a string?

db/seeds.rb
dfaulken: It's seeds, so not super important — just to simulate real data. But I can add a few more if you like.

app/views/passengers/index.haml
dfaulken: https://github.com/umts/st-pax/blob/3f98d3086a213d1de593a907f785ae44a5dea8a7/app/models/passenger.rb#L12

app/views/passengers/index.haml
dfaulken: https://github.com/umts/st-pax/blob/3f98d3086a213d1de593a907f785ae44a5dea8a7/app/views/passengers/index.haml#L1-L3

app/views/passengers/index.haml
mboneil10: OH

db/seeds.rb
mboneil10: Nah, thanks for clarifying.

app/views/passengers/index.haml
mboneil10: I know, but why

dfaulken: That *is* why. So that we can join them. I thought about using symbols, but I'd just need to `map(&:to_s)` them first.

See discussion at #34.

Issue #25, opened 2017-08-21
Set up Codecliamte and Travis


Anbranin:
Need some continuous integration

dfaulken:
> some continuous integration

So Travis too I assume?

Anbranin:
Ya

dfaulken:
https://codeclimate.com/github/umts/st-pax

dfaulken:
All set.

PR #52, opened 2017-08-29
Use travis, and switch to mysql

dfaulken:
Addresses #25.

Note for everyone working on the app that once this is merged you'll need to switch your database as well.

@AkivaGreen @avacmeyer @Chauce


Gemfile
```diff
@@ -2,39 +2,14 @@
...
-git_source(:github) do |repo_name|
-  repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?('/')
-  "https://github.com/#{repo_name}.git"
-end
-
-# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
+gem 'coffee-rails', '~> 4.2'
+gem 'haml'
+gem 'haml-rails'
+gem 'mysql2'
```
dfaulken: Switching `sqlite3` to `mysql2` is the only real change here. Everything else is cleanup/alphabetization.

dfaulken:
Issue now is database permissions on the Travis VM.

My solution for incidents was this:

https://github.com/umts/incidents/blob/23a3f75ccb13b8b02d900a917b730b9196b870ed/.travis.yml#L20

https://github.com/umts/incidents/blob/23a3f75ccb13b8b02d900a917b730b9196b870ed/config/database.travis.yml#L1-L5

Is there a better way? @Anbranin @werebus

Anbranin:
Yeah that's just about what they say to do in the travis documentation
https://docs.travis-ci.com/user/database-setup/#Using-MySQL-with-ActiveRecord
So I think you're good. Although do you need to set a password for a test db?

werebus:
Docs [here](https://docs.travis-ci.com/user/database-setup/#MySQL)  say the following:

1.  You may have to add something to the `services` array in the Travis config
2.  It *should* be username 'travis', blank password.  Note that MySQL differentiates between a blank password and no password.  You might have to put `password: ''` in your `database.yml` ❔  Although the docs don't say that, so...

That said, I have no particular objection to the solution you referenced, either.

dfaulken:
Looks like we need to yarn install in `before_script`? Surprised that these tests include JavaScript.

werebus:
Well, the feature tests render views, right? So, `application.js` needs to be compiled since it's referenced in the application layout, even if the tests don't use that JS.

dfaulken:
Problems at this stage are genuine test failures unrelated to Travis configuration. Merging.

PR #53, opened 2017-08-29
Tests

dfaulken:
Tests pass now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ before_script:
   - cp config/database.yml.travis config/database.yml
   - yarn
 script:
-  - bundle exec rake db:reset
+  - bundle exec rake db:create db:schema:load
   - bundle exec rspec
   - bundle exec codeclimate-test-reporter
 notifications:
--- a/app/assets/stylesheets/passengers.scss
+++ b/app/assets/stylesheets/passengers.scss
@@ -93,7 +93,7 @@ table#key, table#passengers {
   .no_note {
     background-color: #90C3D4;
   }
-  .has_been_overridden{
+  .overridden{
     background-color: #E898FF;
   }
 }
--- a/app/helpers/passengers_helper.rb
+++ b/app/helpers/passengers_helper.rb
@@ -13,8 +13,7 @@ module PassengersHelper
       elsif note.expired?
         'inactive'
       end
-    else
-      'no_note'
+    else 'no_note'
     end
   end
 
--- a/app/helpers/passengers_helper.rb
+++ b/app/helpers/passengers_helper.rb
@@ -6,7 +6,9 @@ module PassengersHelper
     return if passenger.permanent?
     note = passenger.doctors_note
     if note.present?
-      if note.will_expire_within_warning_period?
+      if note.override_expiration?
+        'overridden'
+      elsif note.will_expire_within_warning_period?
         'will_expire_soon'
       elsif note.expired_within_grace_period?
         'expired_within_grace_period'
--- a/app/views/passengers/index.haml
+++ b/app/views/passengers/index.haml
@@ -21,7 +21,7 @@
         %tr
           %td.will_expire_soon Will Expire Within 1 Week
           %td.expired_within_grace_period Expired Within Grace Period
-          %td.has_been_overridden Expiration Overridden by Admin
+          %td.overridden Expiration Overridden by Admin
           %td.inactive Inactive
           %td.no_note No Note
 %p
--- a/spec/factories/doctors_notes.rb
+++ b/spec/factories/doctors_notes.rb
@@ -15,7 +15,7 @@ FactoryGirl.define do
   end
 
   trait :expired_within_grace_period do
-    expiration_date { DoctorsNote.grace_period - rand(3).days }
+    expiration_date { DoctorsNote.grace_period + rand(3).days }
   end
 
   trait :expiring_soon do
@@ -23,6 +23,6 @@ FactoryGirl.define do
   end
 
   trait :expired do
-    expiration_date { DoctorsNote.grace_period + rand(30).days }
+    expiration_date { Date.today + rand(30).days }
   end
 end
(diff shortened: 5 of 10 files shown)


app/assets/stylesheets/passengers.scss
```diff
@@ -93,7 +93,7 @@ table#key, table#passengers {
   .no_note {
     background-color: #90C3D4;
   }
-  .has_been_overridden{
+  .overridden{
```
dfaulken: Seemed more concise.

app/helpers/passengers_helper.rb
```diff
@@ -6,15 +6,16 @@ def passengers_table_row_class(passenger)
     return if passenger.permanent?
     note = passenger.doctors_note
     if note.present?
-      if note.will_expire_within_warning_period?
+      if note.override_expiration?
```
dfaulken: This had been accidentally removed.

spec/factories/doctors_notes.rb
```diff
@@ -15,14 +15,14 @@
...
+    expiration_date { DoctorsNote.grace_period + rand(3).days }
   end
 
   trait :expiring_soon do
     expiration_date { rand(7).days.since }
   end
 
   trait :expired do
-    expiration_date { DoctorsNote.grace_period + rand(30).days }
+    expiration_date { Date.today + rand(30).days }
```
dfaulken: I had misunderstood the way `DoctorsNote.grace_period` works.

.travis.yml
```diff
@@ -8,7 +8,7 @@ before_script:
   - cp config/database.yml.travis config/database.yml
   - yarn
 script:
-  - bundle exec rake db:reset
+  - bundle exec rake db:create db:schema:load
```
dfaulken: Don't want to seed in test.

spec/features/passenger_filters_spec.rb
```diff
@@ -52,8 +52,9 @@
     visit passengers_url
     check('show_inactive')
     click_on('Show Passengers')
-    expect(page).to have_selector('table#passengers tbody tr', count: 1)
+    # Shows both active and inactive passengers
+    expect(page).to have_selector('table#passengers tbody tr', count: 2)
```
mboneil10: This wasn't passing before, correct?

app/assets/stylesheets/passengers.scss
mboneil10: Agreed

spec/features/passenger_filters_spec.rb
dfaulken: Correct.

spec/helpers/passengers_helper_spec.rb
```diff
@@ -37,6 +37,12 @@
         expect(helper.passengers_table_row_class(@passenger)).to eql 'inactive'
       end
     end
+    context 'expiration overridden' do
```
mboneil10: Nice

app/views/passengers/index.haml
```diff
@@ -21,7 +21,7 @@
         %tr
           %td.will_expire_soon Will Expire Within 1 Week
           %td.expired_within_grace_period Expired Within Grace Period
-          %td.has_been_overridden Expiration Overridden by Admin
+          %td.overridden Expiration Overridden by Admin
```
mboneil10: Will there ever be a time when the expiration is overridden by a different user? In other words, do we need to specify that the expiration is overridden by admin?

spec/helpers/passengers_helper_spec.rb
```diff
@@ -37,9 +37,15 @@
...
+      it "returns 'overridden'" do
+        @doctors_note.update override_expiration: true
+        expect(helper.passengers_table_row_class(@passenger)).to eql 'overridden'
+      end
+    end
+    context 'no note' do
       it "returns 'no_note'" do
-        @doctors_note.destroy
+        @passenger = create :passenger, :temporary, :no_note
```
mboneil10: Can you explain this? I could make an educated guess why you made this change, but I just want to be sure.

app/helpers/passengers_helper.rb
```diff
@@ -6,15 +6,16 @@ def passengers_table_row_class(passenger)
...
+        'overridden'
+      elsif note.will_expire_within_warning_period?
         'will_expire_soon'
       elsif note.expired_within_grace_period?
         'expired_within_grace_period'
       elsif note.expired?
```
mboneil10: So if the `note` has expired, that means the `passenger` is `inactive`? And an `inactive passenger` will always have an `expired note`? I'm asking because both return `'inactive'`.

app/views/passengers/index.haml
dfaulken: It's not for specificity, just for clarity, I imagine. Personally I think we might do another `s/Admin/STSPV` to make it even clearer who we're talking about.

But yes, only admins can override expiration dates.

spec/helpers/passengers_helper_spec.rb
dfaulken: I thought destroying the object (my first solution) seemed a little roundabout. Clearer to state "okay now we want the passenger to have no note."

app/helpers/passengers_helper.rb
dfaulken: Correct. Activeness takes precedence, is why you see it here twice. An inactive passenger is inactive regardless of anything to do with their note. An active passenger is shown as inactive if their note is expired (beyond the grace period).

mboneil10: Alright, awesome.

app/views/passengers/index.haml
mboneil10: That sounds like a good idea -- I like STSPV over Admin.

spec/helpers/passengers_helper_spec.rb
mboneil10: Great!

Issue #50, opened 2017-08-29
MobilityDevice#device should be #name


dfaulken:
I would've bundled with #49, but it seemed unrelated and would've added significant bulk.

I'll work on this pending that PR.

PR #54, opened 2017-08-29
Change #device to #name on MobilityDevice

dfaulken:
Pending #53, Travis will tell us whether I did this correctly.

### DIFF
--- a/app/controllers/passengers_controller.rb
+++ b/app/controllers/passengers_controller.rb
@@ -56,7 +56,7 @@ class PassengersController < ApplicationController
   def passenger_params
     permitted_params = params.require(:passenger)
                              .permit :name, :address, :email, :phone,
-                                     :wheelchair, :mobility_device, :active,
+                                     :wheelchair, :mobility_device_id, :active,
                                      :permanent, :note,
                                      doctors_note_attributes: [:expiration_date]
     unless @current_user.admin?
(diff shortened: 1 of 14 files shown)


app/controllers/passengers_controller.rb
```diff
@@ -56,7 +56,7 @@ def destroy
   def passenger_params
     permitted_params = params.require(:passenger)
                              .permit :name, :address, :email, :phone,
-                                     :wheelchair, :mobility_device, :active,
+                                     :wheelchair, :mobility_device_id, :active,
```
dfaulken: Unrelated to this PR, but was definitely a bug. Wouldn't have been possible before to actually update a passenger's mobility device.

dfaulken:
Closes #50.

db/migrate/20170829182335_change_device_to_name_in_mobility_devices.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeDeviceToNameInMobilityDevices < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :mobility_devices, :device, :name
```
mboneil10: This is a fantastic change.

lib/tasks/import/mobility_devices.rake
```diff
@@ -6,7 +6,7 @@ desc 'Import mobility devices from Access CSV at data/mobilitydevice.csv'
 namespace :import do
   task mobility_devices: :environment do
     CSV.foreach('data/mobilitydevice.csv', headers: true).with_index 2 do |row, line|
-      device = MobilityDevice.new device: row.fetch('name')
+      device = MobilityDevice.new name: row.fetch('name')
```
mboneil10: Why didn't you have to change this from `row.fetch('device')` to what it is now?

mboneil10: Ohhhh. Never mind.

PR #56, opened 2017-08-30
Replace admin

### LINKED ISSUE #55: s/Admin/STSPV
Maybe not in the code for this one, but it'd be good to be clear about who 'admins' are.

mboneil10:
Closes #55


app/views/users/show.html.haml
```diff
@@ -11,7 +11,7 @@
   %b Active:
   = @user.active
 %p
-  %b Admin:
+  %b STSPV:
   = @user.admin
```
dfaulken: While we're here, can we change this to not show `true` or `false` in a view?

mboneil10: What do you mean?

mboneil10: I can't find it how do I rails/ruby/haml

dfaulken: I mean that `@user.admin` is the value `true` or `false`. Ideally this would say "Yes" or "No".

PR #57, opened 2017-08-31
Remove phone from users

### LINKED ISSUE #38: Remove phone from User
umasstransit.org is the method through which employees get in contact with one another, I think.

dfaulken:
Closes #38.

### DIFF
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 97.06
+    "covered_percent": 97.05
   }
 }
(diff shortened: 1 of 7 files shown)


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 97.06
+    "covered_percent": 97.05
```
AkivaGreen: How did this go down?

dfaulken: This is because I changed two lines to one the in the users controller.

PR #59, opened 2017-08-31
Add missing attributes

### LINKED ISSUE #27: Add missing attributes to Passenger, and import
The following attributes are present in Access but don't exist in our passengers table:

+ student ID
+ UMass status (student, alumni, etc.)
+ dispatcher who registered them

For temporary passengers:

+ whether they're registered with Disability Services
+ "has brochure" (🤷‍♂️)

Do we need them?

AkivaGreen:
This should resolve #27.


app/assets/javascripts/temp_pax_options.js
```diff
@@ -0,0 +1,18 @@
+$(document).on('turbolinks:load', function(){
+  $('#passenger_permanent').change(function(){
+    var registeredbyField = $('#pax_registered_with_DS');
+    var hasbrochureField = $('#pax_has_brouchure');
```
dfaulken: typo

app/models/passenger.rb
```diff
@@ -7,6 +7,9 @@ class Passenger < ApplicationRecord
                     format: { with: VALID_EMAIL_REGEX }, uniqueness: true
   STATUSES = %w[Alumni Faculty Staff Student].freeze
   validates :status, inclusion: { in: STATUSES, allow_blank: true }
+  validates :spire,
```
dfaulken: Why do we need this?

app/views/passengers/_form.html.haml
```diff
@@ -1,15 +1,21 @@
...
+    %ul
+      - @passenger.errors.full_messages.each do |message|
+        %li= message
+= form_with model: @passenger, local: true do |form|
+  .field
+    = form.label :permanent_passenger
+    = form.check_box :permanent, disabled: @current_user.dispatcher?,
+      id: "passenger_permanent"
   .field
     = form.label :passenger_name
```
dfaulken: Labels and text fields should have matching arguments. If you want to change the text that's shown, use the locales file.

app/views/passengers/_form.html.haml
```diff
@@ -19,22 +25,32 @@
...
+    = form.label :umass_status
+    = form.select :status, Passenger::STATUSES, include_blank: true
+  .field
+    = form.label :registered_with_disability_services
+    = form.check_box :registered_with_disability_services,
+      id: "pax_registered_with_DS"
+  .field
+    = form.label :has_brochure
+    = form.check_box :has_brochure,
+      id: "pax_has_brouchure"
```
dfaulken: typo

app/views/passengers/_form.html.haml
```diff
@@ -19,22 +25,32 @@
...
+  .field
     = form.label :mobility_device_id
     = form.select :mobility_device_id, options_from_collection_for_select(MobilityDevice.order(:name), :id, :name), include_blank: 'None'
   - if @current_user.admin?
     .field
       = form.label :active
       = form.check_box :active
-  .field
```
dfaulken: Why remove this?

app/views/passengers/_form.html.haml
```diff
@@ -19,22 +25,32 @@
...
-    = form.check_box :permanent, disabled: @current_user.dispatcher?,
-      id: "passenger_permanent"
   = form.fields_for :doctors_note, @doctors_note do |d|
     .field
       = d.label :expiration_date
       = d.text_field :expiration_date,
         class: 'datepicker', id: "doctors_note_expiration_date"
   .field
+    = form.label :registering_dispatcher
+    = form.select :registered_by, options_from_collection_for_select(User.all, :id, :name),include_blank: true
```
dfaulken: This shouldn't be in the form. It should be done automatically in the controller when a new passenger is created.

app/views/passengers/show.haml
```diff
@@ -1,14 +1,32 @@
...
   = @passenger.email
 %p
   %b Phone:
   = @passenger.phone
 %p
+  %b UMass Status:
+  = @passenger.status
+%p
+  %b Registered with disability services?
+  - if @passenger.registered_with_disability_services
```
dfaulken: Should use predicate methods for boolean columns.

app/models/passenger.rb
dfaulken: Not just this validation, but SPIRE for passengers in general.

app/views/passengers/show.haml
```diff
@@ -26,7 +44,12 @@
 %p
   %b Expiration Date:
   = @passenger.expiration_display
+%p
+  %b Registering Dispatcher:
+  = @passenger.registered_by
```
dfaulken: Won't this just be an integer?

app/models/passenger.rb
Chauce: I think that was one of the things the ST specifically asked to have in the database. Not sure what the rationale was, though.

app/views/passengers/_form.html.haml
Chauce: I moved it to the top. Checking it causes a couple of boxes that are only relevant to temporary passengers to deactivate, so I thought putting it further up made the most sense.

app/models/passenger.rb
dfaulken: OK.

app/views/passengers/show.haml
```diff
@@ -46,7 +46,7 @@
   = @passenger.expiration_display
 %p
   %b Registering Dispatcher:
-  = @passenger.registered_by
+  = User.where(id: @passenger.registered_by).first.try(:name) || "Unknown"
```
dfaulken: There's an association for this.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 97.05
+    "covered_percent": 72.14
```
Anbranin: :(

app/assets/javascripts/temp_pax_options.js
```diff
@@ -0,0 +1,18 @@
+$(document).on('turbolinks:load', function(){
+  $('#passenger_permanent').change(function(){
+    var registeredbyField = $('#pax_registered_with_DS');
+    var hasbrochureField = $('#pax_has_brochure');
+    if($(this).is(':checked')){
+      registeredbyField.prop('disabled', true);
+      registeredbyField.prop('checked', false);
+    }
+    else registeredbyField.prop('disabled', false)
+         registeredbyField.prop('checked', false)
```
dfaulken: I'm not too sure about the JS here, namely:

```js
else something();
somethingElse();
```

whether both `something`s are actually attached to the `else`.

app/controllers/passengers_controller.rb
```diff
@@ -31,10 +31,10 @@ def index
 
   def create
     @passenger = Passenger.new(passenger_params)
+    @passenger.registered_by = @current_user.id
```
dfaulken: Just `@current_user` is fine.

app/views/passengers/_form.html.haml
```diff
@@ -1,24 +1,41 @@
...
-      %h2= "#{pluralize(@passenger.errors.count, "error")} prohibited this passenger from being saved:"
-      %ul
-        - @passenger.errors.full_messages.each do |message|
-          %li= message
-
-  .field
-    = form.label :passenger_name
+- if @passenger.errors.any?
+  #error_explanation
+    %h2= "#{pluralize(@passenger.errors.count, "error")} prohibited this passenger from being saved:"
```
dfaulken: I like to break with Rails scaffold style here:

```haml
%h2
  = pluralize @passenger.errors.count, 'error'
  prohibited this passenger from being saved:
```

app/views/passengers/_form.html.haml
```diff
@@ -1,24 +1,41 @@
...
-    = form.label :phone_number
+    = form.label :phone
     = form.phone_field :phone
   .field
+    = form.label :status
+    = form.select :status, Passenger::STATUSES, include_blank: true
+  .field
+    = form.label :registered_with_disability_services
+    = form.check_box :registered_with_disability_services,
+      id: "pax_registered_with_DS"
```
dfaulken: These IDs can be symbols.

app/views/passengers/show.haml
```diff
@@ -1,14 +1,32 @@
...
   = @passenger.phone
 %p
+  %b UMass Status:
+  = @passenger.status
+%p
+  %b Registered with disability services?
+  - if @passenger.registered_with_disability_services?
+    Yes
+  - else
+    No
```
dfaulken: It might be nice to translate the `if boolean then 'Yes' else 'No'` into a helper method, or just use the existing one for the cross and check images.

coverage/.last_run.json
dfaulken: This doesn't seem correct. The majority of added code here was in views, which don't factor into test coverage. Do you have test failures or was this perhaps added after running an incomplete test suite?

app/assets/javascripts/temp_pax_options.js
Chauce: It's functional ... I'm checking into whether or not there is a better way to do it. Basically, in the `else` case, I want that particular checkbox to be enabled and unchecked. I guess since it un-checks when the box is disabled, what I'm doing here might actually be redundant since it should be unchecked initially, anyways.

werebus: No, I don't think that syntax is right.  The `'checked'` line is **outside** the else block. It doesn't feel like that what you intended. You probably want to add some `{}`s

Example:   https://jsbin.com/xoqudafaqu/edit?html,js,output

app/assets/javascripts/temp_pax_options.js
```diff
@@ -0,0 +1,22 @@
...
+  $('#passenger_permanent').change(function(){
+    var registeredbyField = $('#pax_registered_with_DS');
+    var hasbrochureField = $('#pax_has_brochure');
+    if($(this).is(':checked')){
+      registeredbyField.prop('disabled', true);
+      registeredbyField.prop('checked', false);
+    }
+    else {
+      registeredbyField.prop('disabled', false)
+      registeredbyField.prop('checked', false)
```
dfaulken: Missing semicolons

app/views/passengers/_form.html.haml
```diff
@@ -1,39 +1,58 @@
...
+      = pluralize @passenger.errors.count, "error"
+      prohibited this passenger from being saved:
+    %ul
+      - @passenger.errors.full_messages.each do |message|
+        %li= message
+= form_with model: @passenger, local: true do |form|
+  .field
+    = form.label :permanent
+    = form.check_box :permanent, disabled: @current_user.dispatcher?,
+      id: "passenger_permanent"
```
dfaulken: This should be a symbol to match the others

coverage/.last_run.json
dfaulken: @Chauce can we re-rerun the specs to verify this percentage?

app/views/passengers/show.haml
```diff
@@ -16,31 +16,19 @@
   = @passenger.status
 %p
   %b Registered with disability services?
-  - if @passenger.registered_with_disability_services?
-    Yes
-  - else
-    No
+  %td.yes= yes_no_image @passenger.registered_with_disability_services?
```
mboneil10: Great job with this! I know you had some trouble to begin with, but you got there, @Chauce!

app/assets/javascripts/temp_pax_options.js
```diff
@@ -0,0 +1,22 @@
...
+    var registeredbyField = $('#pax_registered_with_DS');
+    var hasbrochureField = $('#pax_has_brochure');
+    if($(this).is(':checked')){
+      registeredbyField.prop('disabled', true);
+      registeredbyField.prop('checked', false);
+    }
+    else {
+      registeredbyField.prop('disabled', false);
+      registeredbyField.prop('checked', false);
+    }
```
dfaulken: You could boil down this conditional down to:

```js
registeredbyField.prop('disabled', $(this).is(':checked'));
registeredbyField.prop('checked', false);
```

and the same for the other field.

Chauce: Without the if/else? I've been trying to figure out how to re-write these lines but I'm doing something wrong ...

dfaulken: Correct. You can completely replace the conditional with those two lines.

`$(this).is(':checked')` | disabled | checked
--- | --- | ---
true | true | false
false | false | false

So you can always set `'checked'` to false and you always set `'disabled'` to match `$(this).is(':checked')`.

Chauce:
Also closes #69

PR #61, opened 2017-08-31
Reactivation warning

### LINKED ISSUE #9: Enhance the reactivation warning
Current functionality: When a user checks the "active" button to reactivate a passenger, there is a javascript popup warning them that if the expiration date is not also updated the passenger will become deactivated again at 4:30 AM.

Problems: If a warning pops up every time a passenger is reactivated, alert fatigue can occur. Alerts when something is not actually wrong or going to go wrong is undesirable.

Desired functionality: Write a Javascript function that checks the expiration date value in the field and whether or not it is in the future. If the date is in the future, no warning. If the date is today or before, warn the user. That way, they only get a warning if they have done something that may cause problems in the future, thus preventing alert fatigue.

AkivaGreen:
This should resolve #9.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -17,3 +17,4 @@
 //= require rails-ujs
 //= require turbolinks
 //= require_tree .
+//= require moment/moment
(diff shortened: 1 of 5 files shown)


dfaulken:
Can we see what this looks like?

AkivaGreen:
screen shot 2017-08-31 at 2 28 12 pm

AkivaGreen:
screen shot 2017-08-31 at 2 30 42 pm

PR #80, opened 2017-09-14
Asset fixes

werebus:
I think I found all the places where various asset libraries were circumvrenting the pipeline. These are all things that work in development, but don't work if you precompile assets.


werebus:
To be fair, I think we're not actually *using* most of those UI elements. We could take a bunch of them out, but I wanted to play it safe and just replicate what was there before.

Issue #48, opened 2017-08-28
Interface to override expiration date


dfaulken:
For admins.

Needed for deployment.

dfaulken:
@Chauce it seems natural for you to roll this into #27. Would you mind taking this on also while you're doing that?

Chauce:
I can give @jforgue a hand if need be. The only thing blocking me was what the interface should look like.

jforgue:
Closed with PR #85

PR #85, opened 2017-09-20
Create UI for overriding expiration date

jforgue:


app/controllers/passengers_controller.rb
```diff
@@ -40,6 +41,7 @@ def create
 
   def update
     if @passenger.update(passenger_params)
+      @passenger.doctors_note.update_attributes(overridden_by: @current_user)
```
dfaulken: We don't want to do this every time someone saves a passenger, right? Only when someone's specifically in the process of overriding the doctor's note.

app/helpers/passengers_helper.rb
```diff
@@ -26,4 +26,8 @@ def passengers_table_class
       'row-border dispatch-table'
     end
   end
+
+  def passengers_form
```
dfaulken: It'd be nice if this method indicates that what it returns is a CSS class.

app/models/doctors_note.rb
```diff
@@ -2,6 +2,13 @@
 
 class DoctorsNote < ApplicationRecord
   belongs_to :passenger
+  belongs_to :overridden_by, class_name: 'User', required: false
+  before_save :check_override
+
+  def check_override
```
dfaulken: Should be a private method.

app/views/passengers/_form.html.haml
```diff
@@ -48,9 +51,18 @@
     .field
       = d.label :expiration_date
       = d.text_field :expiration_date,
-        class: 'datepicker', id: :doctors_note_expiration_date
+        class: 'datepicker', id: 'doctors_note_expiration_date'
```
dfaulken: Why this change? Symbol should be fine here.

app/views/passengers/_form.html.haml
```diff
@@ -48,9 +51,18 @@
...
+    .field
+      = d.label :override_expiration_date
```
dfaulken: `override_expiration_date` isn't the name of an attribute in the model, right?

app/views/passengers/_form.html.haml
```diff
@@ -48,9 +51,18 @@
...
+      = d.check_box :override_expiration,
+        id: 'check-click'
+    - if @doctors_note.present?
+      .field#slide-date{class: passengers_form}
+        = d.label :enter_override_date
```
dfaulken: Same here. `enter_override_date` isn't the name of a model attribute.

app/views/passengers/_form.html.haml
```diff
@@ -48,9 +51,18 @@
...
+        = d.text_field :override_until,
+          class: 'datepicker'
   .field
-    = form.label "Additional Notes"
+    = form.label :additional_notes
```
dfaulken: Same here too.

db/migrate/20170913210806_change_overridden_by_to_overridden_by_id.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeOverriddenByToOverriddenById < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :doctors_notes, :overridden_by, :overridden_by_id
```
dfaulken: I disagree with this change. It's pretty common to have `_by` columns which point to users. Instead, you should tell the model that it's a foreign key. There are examples in our other apps.

spec/factories/doctors_notes.rb
```diff
@@ -9,8 +9,7 @@
 
   trait :overridden do
     override_expiration true
-    # TODO: I think this should be defined as a real association.
-    overridden_by 1
+    overridden_by { create :user, :admin }
```
dfaulken: 👍

app/models/doctors_note.rb
jforgue: Right, I knew that.

app/views/passengers/_form.html.haml
```diff
@@ -49,20 +49,20 @@
...
     - if @doctors_note.present?
-      .field#slide-date{class: passengers_form}
-        = d.label :enter_override_date
+      .field#slide-date{class: passengers_form_class}
+        = d.label 'enter_override_date'
         = d.text_field :override_until,
           class: 'datepicker'
   .field
-    = form.label :additional_notes
+    = form.label 'additional_notes'
```
dfaulken: No, symbols are correct here. They should match the names of the columns of the model.

dfaulken: In this case `:note`, and above `:override_expiration`

spec/helpers/passengers_helper_spec.rb
```diff
@@ -21,7 +21,8 @@
     context 'will expire within warning period' do
       it "returns 'will_expire_soon'" do
         @doctors_note.update expiration_date: 5.days.since
-        expect(helper.passengers_table_row_class(@passenger)).to eql 'will_expire_soon'
+        expect(helper.passengers_table_row_class(@passenger))
```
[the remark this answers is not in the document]
mboneil10: Or did you write these tests?

app/views/passengers/_form.html.haml
jforgue: Okay I see what you mean. I didn't realize you could change what would be displayed in the en.yml.

spec/helpers/passengers_helper_spec.rb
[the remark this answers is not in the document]
jforgue: I only enforced rubocop on my changes and let rubocop -a fix what it could elsewhere.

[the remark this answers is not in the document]
mboneil10: Alright, no problem with a small app like this one. 👍

app/helpers/passengers_helper.rb
```diff
@@ -26,4 +26,8 @@ def passengers_table_class
       'row-border dispatch-table'
     end
   end
+
+  def passengers_form_class
+    'hide_view' unless @doctors_note.override_expiration?
```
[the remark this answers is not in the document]
mboneil10: Is there a reason why you chose to do this here instead of in the HAML? Just wondering.

jforgue:
Closes #48

spec/helpers/passengers_helper_spec.rb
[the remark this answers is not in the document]
jforgue: I opened an issue to finish testing. 👍

app/helpers/passengers_helper.rb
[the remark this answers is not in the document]
jforgue: I felt this kind of logic should stay out of the view and the helper was already there. 👍

dfaulken:
Sorry for the delay on this one @jforgue. Great work.

PR #86, opened 2017-09-24
Style links and buttons

### LINKED ISSUE #83: Link is Gray but not Disabled
I noticed that the `visited links` are [gray](https://github.com/umts/st-pax/blob/master/app/assets/stylesheets/application.scss#L61). Whenever I see _anything_ on a page that is grayed out, I automatically think it's disabled. Compare the top two passengers' `view` and `delete` to the bottom one:
screen shot 2017-09-17 at 10 18 19 am

The reason I bring this up is because we have gray-on-gray:
screen shot 2017-09-17 at 10 28 28 am

And that same color indicates that a button is disabled:
screen shot 2017-09-17 at 10 21 41 am

My fix would be to remove styling on visited links. Or maybe remove disabled buttons from the user. Possibly both. Could be a personal problem?

mboneil10:
Closes #83

### DIFF
--- a/app/views/passengers/_form.html.haml
+++ b/app/views/passengers/_form.html.haml
@@ -30,9 +30,10 @@
     = form.label :status
     = form.select :status, Passenger::STATUSES, { include_blank: true }, { id: :passenger_status }
   .field
-    = form.label :registered_with_disability_services
-    = form.check_box :registered_with_disability_services,
-      id: :pax_registered_with_DS, disabled: @passenger.permanent?
+    - unless @passenger.permanent?
+      = form.label :registered_with_disability_services
+      = form.check_box :registered_with_disability_services,
+        id: :pax_registered_with_DS, disabled: @passenger.permanent?
   .field
     = form.label :has_brochure
     = form.check_box :has_brochure,
--- a/app/views/passengers/_form.html.haml
+++ b/app/views/passengers/_form.html.haml
@@ -35,9 +35,10 @@
       = form.check_box :registered_with_disability_services,
         id: :pax_registered_with_DS, disabled: @passenger.permanent?
   .field
-    = form.label :has_brochure
-    = form.check_box :has_brochure,
-      id: :pax_has_brochure, disabled: @passenger.permanent?
+    - unless @passenger.permanent?
+      = form.label :has_brochure
+      = form.check_box :has_brochure,
+        id: :pax_has_brochure, disabled: @passenger.permanent?
   .field
     = form.label :mobility_device_id
     = form.select :mobility_device_id, options_from_collection_for_select(MobilityDevice.order(:name), :id, :name), include_blank: 'None'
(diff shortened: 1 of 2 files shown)


mboneil10:
Probably safe to remove disabling the checkbox logic?

app/views/passengers/_form.html.haml
```diff
@@ -9,8 +9,9 @@
 = form_with model: @passenger, local: true do |form|
   .field
     = form.label :permanent
-    = form.check_box :permanent, disabled: @current_user.dispatcher?,
-      id: :passenger_permanent
+    - unless @current_user.dispatcher?
+      = form.check_box :permanent, disabled: @current_user.dispatcher?,
```
Anbranin: Why disable it if the current user is a dispatcher while in a block that won't run if the current user is a dispatcher?

app/views/passengers/_form.html.haml
```diff
@@ -30,13 +31,15 @@
     = form.label :status
     = form.select :status, Passenger::STATUSES, { include_blank: true }, { id: :passenger_status }
   .field
-    = form.label :registered_with_disability_services
-    = form.check_box :registered_with_disability_services,
-      id: :pax_registered_with_DS, disabled: @passenger.permanent?
+    - unless @passenger.permanent?
+      = form.label :registered_with_disability_services
+      = form.check_box :registered_with_disability_services,
+        id: :pax_registered_with_DS, disabled: @passenger.permanent?
```
Anbranin: Same comment here.

app/views/passengers/_form.html.haml
dfaulken: This is what @mboneil10 meant by

> Probably safe to remove disabling the checkbox logic?

app/views/passengers/_form.html.haml
```diff
@@ -30,13 +31,15 @@
...
   .field
-    = form.label :has_brochure
-    = form.check_box :has_brochure,
-      id: :pax_has_brochure, disabled: @passenger.permanent?
+    - unless @passenger.permanent?
+      = form.label :has_brochure
+      = form.check_box :has_brochure,
+        id: :pax_has_brochure, disabled: @passenger.permanent?
```
Anbranin: I'm sensing a pattern here... am I wrong about this?

app/views/passengers/_form.html.haml
Anbranin: I see now! Yes.

app/views/passengers/_form.html.haml
mboneil10: Nope. It's up to you and @chauce if I should hide these or just keep it the way it was (disabling boxes).

Anbranin: My personal preference is that the box is just disabled but I feel not strongly at all about this. I can just imagine maybe _maybe_ some confusion on the part of the dispatcher at some point "hey, don't I have to check if this person is registered or has a brochure?" but it's really whatever.

mboneil10: Ok. That second part is what I was thinking about, too.

PR #90, opened 2017-09-29
Disable perm. passenger fields on page load 

jforgue:
This PR also adds a check for the presence of a passengers note before updating passenger_note_params.

### DIFF
--- a/app/controllers/passengers_controller.rb
+++ b/app/controllers/passengers_controller.rb
@@ -41,7 +41,7 @@ class PassengersController < ApplicationController
 
   def update
     @passenger.assign_attributes passenger_params
-    if @passenger.doctors_note.override_until_changed? && @current_user.admin?
+    if @passenger.doctors_note.present? && @passenger.doctors_note.override_until_changed? && @current_user.admin?
       @passenger.doctors_note.assign_attributes overridden_by: @current_user
     end
     if @passenger.save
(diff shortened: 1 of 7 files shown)


app/controllers/passengers_controller.rb
```diff
@@ -41,7 +41,7 @@ def create
 
   def update
     @passenger.assign_attributes passenger_params
-    if @passenger.doctors_note.override_until_changed? && @current_user.admin?
+    if @passenger.doctors_note.present? && @passenger.doctors_note.override_until_changed? && @current_user.admin?
```
dfaulken: I think this could be more concisely handled with a `try`.

spec/features/passenger_management_spec.rb
```diff
@@ -65,7 +65,7 @@
       click_link 'Add New Passenger'
       fill_in('passenger[name]', with: 'Foo Bar')
       fill_in('passenger[email]', with: 'foobar@invalid.com')
-      fill_in('doctors_note_expiration_date', with: date)
+      fill_in('passenger[doctors_note_attributes][expiration_date]', with: date)
```
dfaulken: It would be good, perhaps in a separate PR, to fix the form so that the view tests can work properly. Example: https://github.com/umts/st-pax/pull/81/files

Issue #58, opened 2017-08-31
Weird error pops up on logout


AkivaGreen:
It still logs you out, but...

![screen shot 2017-08-31 at 11 33 26 am](https://user-images.githubusercontent.com/15308734/29931803-4c6de882-8e40-11e7-807b-8ca3ff735243.png)

dfaulken:
Probably a too generic invocation of the datatables logic in the Javascript. We should only be trying to create a datatable if we're on the passengers index page, not turbolinking to the logout page.

werebus:
So, according to the page referenced in the error message.  This is caused by reinitializing `DataTables` with **different** options to the existing DataTable.

I think this is related to the admin/non-admin invocation.  [Here][1], and then again [here][2].  Is there a chance that loging out causes that div to have *both* the `.admin_table` *and* the `.dispatch_table` class?

Chauce:
Branch `logout-error` contains the fix for this now :)

Chauce:
Fix for this will likely also close #96

Chauce:
<img width="585" alt="screen shot 2018-06-20 at 1 12 22 pm" src="https://user-images.githubusercontent.com/19393198/41673683-a2494ef4-748b-11e8-84b0-52b32d08b272.png">
It's _possible_ that the fix for #96 will **not** close this issue. I think the cause of this issue is that we have a dispatcher table and an admin table, and once you log out, the site is briefly trying to initialize both.

PR #91, opened 2017-10-04
Duplicate header footer

### LINKED ISSUE #71: Navigating "back" to passengers index creates duplicate data table headers and footers
1.  Load the `PassengersController#index`
2.  Go somewhere else
3.  Press the back button
4.  Repeat steps 2 and 3 if you like.

screenshot_2017-09-13_13-27-31
screenshot_2017-09-13_13-27-50

Yet another turbolinks/datatables conflict of some kind?  I'd be curious if #65 might fix this too?

Chauce:
Closes #71


app/assets/javascripts/datatables.coffee
```diff
@@ -0,0 +1,24 @@
...
+  ]
+  i = 0
+  while i < tableElementIds.length
+    tableElementId = tableElementIds[i]
+    if $.isEmptyObject($.find(tableElementId))
+      i++
+      continue
+    table = undefined
+    if $.fn.DataTable.isDataTable(tableElementId)
+      table = $(tableElementId).DataTable()
```
werebus: Seems like this would cause the same error as #58, right?

app/assets/javascripts/datatables.coffee
```diff
@@ -0,0 +1,24 @@
...
+    else
+      table = $(tableElementId).DataTable() #Some options would go here...
```
werebus: Oh, would they?  What sort of options?

app/assets/javascripts/datatables.coffee
```diff
@@ -0,0 +1,24 @@
+$(document).on 'turbolinks:load', ->
+  tableElementIds = [
+    '#passengers.admin-table'
+    '#passengers.dispatch-table'
+  ]
+  i = 0
+  while i < tableElementIds.length
```
werebus: Seems odd to use a while loop and an array rather than jQuery's selectors and each.

app/assets/javascripts/datatables.coffee
```diff
@@ -0,0 +1,24 @@
...
+    tableElementId = tableElementIds[i]
+    if $.isEmptyObject($.find(tableElementId))
```
werebus: Aren't these IDs _always_ on the page? any reason to expect `$.find` not to find them?

app/assets/javascripts/datatables.coffee
Chauce: I think the coffeescript can theoretically handle what datatables.js does (choosing whether the admin table or dispatch table shows up). But ... well the js is handling that nicely right now.

Chauce:
So I based this off a coffeescript fix I found for the Datatables plugin on the Datatables help forum. I have tried trimming parts of the code (only using lines 18-20, only 1-5 + 18-20) and it just breaks the Datatables plugin. I am ... at a loss.

app/assets/javascripts/datatables.coffee
Chauce: Here's the weird thing ... If I get rid of my `else` statement but leave the `if`, the duplicate headers/footers error returns. So that means ... `DataTable.isDataTable(tableElementId)` is false? (Or at least, not true?) It's not a DataTable tableElementId? I'm trying to figure out how to parse this and what the implications are ...

Anbranin:
@Chauce this PR is pretty stale..

Chauce:
As soon as Service Notes is all set, finishing this is the next priority

Anbranin:
excellent

werebus:
This ended up being resolved by #126

Issue #92, opened 2017-10-13
Interface to create PDF of current temporary passengers


Anbranin:
The Dispatchers are used to having a printout of temporary passengers they can quickly check when people call in.

We should create a PDF of these - their names and the expiration dates of their doctor's notes in separate columns.

The passengers whose doctors notes are expiring soon (3 days? we can check with Andrew) should be highlighted in some way (let's try not to do colors, maybe shading?)

dfaulken:
Isn't a listing of temporary passengers the point of the website?

Anbranin:
Yes. Andrew came in and requested it. *shrug*

Anbranin:
Could talk to him.

werebus:
:zap: Avia asked about this at the staff meeting yesterday.

Also, we *did* have a conversation with ST about why this is needed. The rationale was that **ROs** need to see someone's temporary status, and they don't have logins to the site. We could go the route of making the site usable by them, but that might be of limited utility for read-only access to a list of people.

dfaulken:
> ROs need to see someone's temporary status, and they don't have logins to the site

IMO this might be a good use case for developing an API between this site and umasstransit.org, then, given that dispatchers are people who are by definition already using another of our websites.

But in the short term, agreed that a PDF is the path of least resistance.

jforgue:
Is this something that should be handled with `prawn`?

Anbranin:
I'd choose prawn based on my familiarity with it, why, did you have another way to generate PDFs you wanted to try?

jforgue:
No, I've just never done this before.

Anbranin:
Assigning this to @Chauce because they need something to work on but @jforgue if you're interested, feel free to pair.

PR #94, opened 2018-04-17
Pdf for st

avacmeyer:
closing #92  
screen shot 2018-04-17 at 3 24 55 pm


app/controllers/passengers_controller.rb
```diff
@@ -27,6 +27,13 @@ def index
       @passengers = @passengers.active
       @filters << 'active'
     end
+
+    if( params[:print].present? )
```
Anbranin: You don't need these parentheses, and I think they clutter a little.

app/controllers/passengers_controller.rb
```diff
@@ -27,6 +27,13 @@ def index
...
+      pdf = PassengersPDF.new(@passengers, @filters)
+      send_data pdf.render, filename: "Going to be the date",
```
Anbranin: Is this the final filename you wanted for this PDF?

app/pdfs/passengers_PDF.rb
```diff
@@ -0,0 +1,34 @@
+require 'prawn/table'
+
+class PassengersPDF < Prawn::Document
+  def initialize(passengers, filters)
+    super(page_layout: :landscape, page_size: "A3" )
```
Anbranin: 1. Does this `page_layout: :landscape` deal actually work? Because I saw the PDF and the page layout seems to be `:letter`. 
2. What does `page_size: "A3" do?

app/pdfs/passengers_PDF.rb
```diff
@@ -0,0 +1,34 @@
...
+    content_width = bounds.width - 10
+    header(filters)
+    mycooltable(passengers)
+  end
+
+  def mycooltable(passengers)
```
Anbranin: Did you want to keep the name `mycooltable` for this method? Hint: no

app/pdfs/passengers_PDF.rb
```diff
@@ -0,0 +1,34 @@
...
+      passenger_table << [ name, mobility_device, phone, expiration, note ]
+    end
+    table(passenger_table) do
+      row(0).font_style = :bold
+    end
+  end
+
+  def header(filters)
+    font_size 30
+    date = Time.now.strftime("%m/%d/%Y").to_s
```
Anbranin: `strftime` already takes a date and turns it into a string. Calling `.to_s` on it is redundant.

app/pdfs/passengers_PDF.rb
werebus: 1. Yes, it does work.
2. http://www.papersizes.org/a-paper-sizes.htm. But don't use that, 11x17 is `:tabloid`

werebus: Sorry, that should be `'TABLOID'`

app/pdfs/passengers_PDF.rb
```diff
@@ -0,0 +1,34 @@
+require 'prawn/table'
+
+class PassengersPDF < Prawn::Document
```
werebus: This filename is incorrect for autoload to work on non-Macs.

Rails expects this class to be in a file named `passengers_pdf.rb` (or `passengers_p_d_f.rb`, but don't) Mac has a dumb, case-insensitive filesystem, which makes it work, but this won't work in production.

app/controllers/passengers_controller.rb
avacmeyer: Yes..... Is that bad?

app/pdfs/passengers_PDF.rb
avacmeyer: lol yeah i'll change it up

werebus:
On the subject of testing PDFs, you can do this with a feature test and the `pdf-inspector` [gem](https://github.com/prawnpdf/pdf-inspector):

```ruby
    visit passengers_path
    click_on 'Print This Page'
    analysis = PDF::Inspector::Text.analyze(page.body)
    expect(analysis.strings).to include 'Something That Should Be in the PDF'
```

app/pdfs/passengers_PDF.rb
```diff
@@ -0,0 +1,34 @@
...
+  def initialize(passengers, filters)
+    super(page_layout: :landscape, page_size: "A3" )
+    content_width = bounds.width - 10
```
werebus: I think that this doesn't do anything. It just sets a local variable while initializing, but that variable is discarded.

app/views/passengers/index.haml
```diff
@@ -14,6 +14,7 @@
     %label{ for: 'show_inactive' } Show Inactive
     = check_box_tag :show_inactive, true, !@filters.include?('active')
   = submit_tag 'Show Passengers'
+  = submit_tag 'Print This Page', name: "print"
```
werebus: Small thing: this button doesn't print "this page" it prints what you set the filters to.

app/pdfs/passengers_PDF.rb
```diff
@@ -0,0 +1,34 @@
...
+    header(filters)
+    mycooltable(passengers)
+  end
+
+  def mycooltable(passengers)
+    font_size 14
+    passenger_table = [["Name", "Mobility Device ", "Phone", "Expiration Date", "Notes"]]
+
+    passengers.each do |passenger|
+      name = passenger.name || ''
```
werebus: I don't think you need all of these `|| ''`s.  That is, doesn't the table cell look the same if you give it `nil` instead of `''`?

app/pdfs/passengers_PDF.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+    table(passenger_table) do
+      row(0).font_style = :bold
+    end
+  end
+
+  def header(filters)
+    font_size 30
+    date = Time.now.strftime("%m/%d/%Y").to_s
+    text filters.map(&:capitalize).join(' ') + " " + "Passengers" + " " + date, style: :bold, size: 30
```
werebus:
```ruby
text (filters + ['Passengers', date]).map(&:capitalize).join(' '), style: :bold, size: 30
```

perhaps?

app/pdfs/passengers_PDF.rb
werebus: At least as long ago as 0.0.2 ( the oldest version on GitHub, c. 2014) [it was fine to use `nil`](https://github.com/prawnpdf/prawn-table/blob/dc9aa9bc2a85931739f40fbcad51b69ffea0ce62/lib/prawn/table/cell.rb#L164-L165). So... the last thing?

app/controllers/passengers_controller.rb
Anbranin: You want the filename to be "Going to be the date" ?

app/controllers/passengers_controller.rb
```diff
@@ -28,9 +28,9 @@ def index
       @filters << 'active'
     end
 
-    if( params[:print].present? )
+    if params[:print].present?
       pdf = PassengersPDF.new(@passengers, @filters)
-      send_data pdf.render, filename: "Going to be the date",
+      send_data pdf.render, filename: "Current Temporary Passengers",
```
Anbranin: Since people can print passengers other than the temporary passengers, why don't you just put whatever the filter is in here? Like
``` ruby
"#{@filters.join(' ') Passengers as of #{Date.today}"
```

app/pdfs/passengers_pdf.rb
```diff
@@ -0,0 +1,34 @@
+require 'prawn/table'
+
+class PassengersPDF < Prawn::Document
+  # what is the name of this file?
+  def initialize(passengers, filters)
+    super(page_layout: :landscape, page_size: "TABLOID" )
```
mboneil10: picky, but you don't need the space before the right parenthesis.

app/pdfs/passengers_pdf.rb
```diff
@@ -0,0 +1,34 @@
...
+    header(filters)
+    passengers_table(passengers)
+  end
+
+  def passengers_table(passengers)
+    font_size 14
+    passenger_table = [["Name", "Mobility Device ", "Phone", "Expiration Date", "Notes"]]
```
mboneil10: Maybe I'm being silly, but `"Mobility Device "` probably doesn't need the space after it.

spec/features/passenger_filters_spec.rb
```diff
@@ -57,4 +57,14 @@
     expect(page).to have_content 'Mary Sue'
     expect(page).to have_content 'Gary Stue'
   end
+  scenario 'Printing a PDF' do
+    visit passengers_url
```
mboneil10: Make sure to update these tests!

PR #101, opened 2018-08-29
Use utf8mb4 as the default encoding

werebus:
utf8 is garbage. 

See https://medium.com/@adamhooper/in-mysql-never-use-utf8-use-utf8mb4-11761243e434


Anbranin:
I usually let other people merge their own PRs but muscle memory absolutely took over and it was done before I realized it was happening.

PR #104, opened 2018-09-14
Set active default on passengers

### LINKED ISSUE #102: New passengers created by dispatchers default to "inactive"
Because dispatchers don't get to see the active/inactive checkbox, new passengers that they create are inactive (because the parameter isn't there) and the dispatcher also can't _make_ them active.

dfaulken:
Closes #102.

Don't know why there was a collation difference.


dfaulken:
Thank you! Good to know.

PR #106, opened 2018-12-04
Force 'permanent' and 'active' to be true or false

werebus:
Also fixed the model scope which shouldn't technically be required given
the schema change. And, changed the form elements to be disabled rather
than hidden to give users a bit more of a hint of what's going on.


db/migrate/20181204180756_change_active_and_permanent_to_not_null.rb
```diff
@@ -0,0 +1,7 @@
+class ChangeActiveAndPermanentToNotNull < ActiveRecord::Migration[5.1]
+  def change
+    change_column_null :passengers, :active, false, true
```
werebus: Note that I'm assuming that passengers with a `NULL` active state should be made active. That _might_ not be what we want. Any thoughts?

Anbranin: Presumably people didn't _deactivate_ them, they just created them, so I think it will depend and we should send a list of the passengers this affects to the ST supervisor like "hey we just activated these passengers: " that way they can do something about it.

werebus: Whew. There sure are 45 of them:

```
Passenger.where(active: nil).count
   (1.2ms)  SELECT COUNT(*) FROM `passengers` WHERE `passengers`.`active` IS NULL
=> 45
```

werebus: I'm going to get Tiff and Avia a list

werebus:
Waiting on input from ST

werebus:
Heard back from ST, `active: nil` should be `active: false`.  @Anbranin, still look OK to you?

PR #125, opened 2019-09-24
Remove override

### LINKED ISSUE #122: Remove all logic related to "override expiration date"
No longer necessary because of #120 and the ability to edit when a doctor's note expired. And no longer desired by STSPVs.

Anbranin:
closes #122


Anbranin:
> test coverage dropped. Can you fix that?

So this is a good observation--when you either remove tests or add new code, test coverage will drop and thus one should write tests to make up for it. However, I neither added code nor removed tests that covered anything other than the code I removed.

The coverage in my case actually dropped because of how percentages work, and isn't in this case indicative of a problem. A peek at `coverage/index.html` tells me, for example, that coverage for the doctor's note model went from 95.65 % to 94.74 %. There's only one line that's uncovered in that entire model, but that one line equals a larger percentage of the entire model code because I removed other lines.

PR #126, opened 2019-09-25
Datatables fix

### LINKED ISSUE #58: Weird error pops up on logout

### LINKED ISSUE #96: Cannot reinitialize DataTable
Navigating "back" to passengers index causes an error message to occur.

> DataTables warning: table id=passengers - Cannot reinitialise DataTable. For more information about this error, please see [http://datatables.net/tn/3](http://datatables.net/tn/3)

Looks like a relatively easy fix.

Anbranin:
closes #96 
closes #58 

I didn't set out to fix these issues, but as I was working on another PR I got so pissed off about them that I got obsessed. It also got in the way of what I was working on, given that I'm making another page that includes a datatable.

### DIFF
--- a/app/assets/javascripts/datatables.js
+++ b/app/assets/javascripts/datatables.js
@@ -1,32 +1,6 @@
 $( document ).on("turbolinks:load", function() {
-  $('#passengers.admin-table').DataTable({
-    "aoColumns": [
-      null,
-      null,
-      null,
-      { "bSortable": false },
-      { "bSortable": false },
-      null,
-      { "bSortable": false },
-      { "bSortable": false },
-      { "bSortable": false },
-    ],
-    "aLengthMenu": [[25, 50, 75, 100, -1], [25, 50, 75, 100, "All"]],
-    "iDisplayLength": 100
-  });
-
-  $('#passengers.dispatch-table').DataTable({
-    "aoColumns": [
-      null,
-      null,
-      null,
-      { "bSortable": false },
-      { "bSortable": false },
-      null,
-      { "bSortable": false },
-      { "bSortable": false },
-    ],
-    "aLengthMenu": [[25, 50, 75, 100, -1], [25, 50, 75, 100, "All"]],
-    "iDisplayLength": 100
+  $('.passengers').DataTable({
+    destroy: true,
+    lengthMenu: [ [10, 25, 50, -1], [10, 25, 50, 'All'] ]
   });
 });
--- a/app/assets/javascripts/datatables.js
+++ b/app/assets/javascripts/datatables.js
@@ -1,5 +1,5 @@
 $( document ).on("turbolinks:load", function() {
-  $('.passengers').DataTable({
+  $('#passengers').DataTable({
     destroy: true,
     lengthMenu: [ [10, 25, 50, -1], [10, 25, 50, 'All'] ]
   });
--- a/app/helpers/passengers_helper.rb
+++ b/app/helpers/passengers_helper.rb
@@ -19,11 +19,9 @@ module PassengersHelper
     doctors_note_row_class(passenger.doctors_note)
   end
 
-  def passengers_table_class
-    if @current_user.admin?
-      'row-border admin-table'
-    else
-      'row-border dispatch-table'
+  def sortable_date(note)
+    if note.present?
+      note.expiration_date.strftime('%Y%m%d%H%M%s')
     end
   end
 end
--- a/app/helpers/passengers_helper.rb
+++ b/app/helpers/passengers_helper.rb
@@ -21,7 +21,7 @@ module PassengersHelper
 
   def sortable_date(note)
     if note.present?
-      note.expiration_date.strftime('%Y%m%d%H%M%s')
+      note.expiration_date.strftime('%Y%m%d')
     end
   end
 end
--- a/app/views/passengers/index.haml
+++ b/app/views/passengers/index.haml
@@ -15,7 +15,7 @@
     = check_box_tag :show_inactive, true, !@filters.include?('active')
   = submit_tag 'Show Passengers'
   = submit_tag 'Print These Filters', name: "print"
-  .checkbox-inline.key_position
+  .key_position
     %table#key
       %tbody
         %tr
--- a/app/views/passengers/index.haml
+++ b/app/views/passengers/index.haml
@@ -31,9 +31,9 @@
         %th Email
         %th Phone
         - unless @filters.include? 'Active'
-          %th Active
-        %th Permanent
-        %th Expiration Date
+          %th.text-center Active
+        %th.text-center Permanent
+        %th.text-center Expiration Date
         %th{data: {orderable: 'false'}}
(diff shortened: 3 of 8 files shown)


app/views/passengers/index.haml
```diff
@@ -24,7 +24,7 @@
           %td.inactive Inactive
           %td.no_note No Note
 - if @passengers.present?
-  %table#passengers{class: passengers_table_class}
+  %table.passengers
```
werebus: Any particular reason for the change from id to class?

app/views/passengers/index.haml
```diff
@@ -15,7 +15,7 @@
     = check_box_tag :show_inactive, true, !@filters.include?('active')
   = submit_tag 'Show Passengers'
   = submit_tag 'Print These Filters', name: "print"
-  .checkbox-inline.key_position
+  .key_position
```
werebus: Minor nit-pick: I'm not sure why this `div` is necessary. The `table` itself can be positioned just as easily.

app/views/passengers/index.haml
werebus: In particular, there's still CSS that targets `table#passengers {`

app/helpers/passengers_helper.rb
```diff
@@ -19,11 +19,9 @@ def passengers_table_row_class(passenger)
...
   end
 
-  def passengers_table_class
-    if @current_user.admin?
-      'row-border admin-table'
-    else
-      'row-border dispatch-table'
+  def sortable_date(note)
+    if note.present?
+      note.expiration_date.strftime('%Y%m%d%H%M%s')
```
werebus: Perhaps a little precise for an expiration "date".  Datatables supports "secondary sort", but it doesn't work if two things that _look_ identical are actually different.

app/views/passengers/index.haml
```diff
@@ -49,7 +49,7 @@
           - unless @filters.include? 'Active'
             %td.yes= yes_no_image passenger.active?
           %td.yes= yes_no_image passenger.permanent?
-          %td.yes= passenger.expiration_display
+          %td.yes{data: {sort: sortable_date(passenger.doctors_note)}}=passenger.expiration_display
```
werebus: On the one-hand, I asked myself why this column has a class that _appears_ to be only for yes-no icons.

On the other hand, that class turns out to not really have anything to do with yes-no icons, it just makes the text centered :roll_eyes: (As does the identical, unused, class `.no`)

On the third hand, why are expiration dates centered when no other text is?

app/views/passengers/index.haml
Anbranin: I guess since there was more than just one now that I've added (in another PR but still) the archived passenger page that uses the same kind of table, I felt weird about having it be an ID. BUT, that rule may in fact only apply to divs on the same page as each other. So I'll change it back (and not changing the css was a massive oversight).

app/helpers/passengers_helper.rb
Anbranin: That's what I get for copying and pasting off the internet.

app/views/passengers/index.haml
Anbranin: > why are expiration dates centered when no other text is?

because it looks goddamn weird.
Screen Shot 2019-09-26 at 11 16 33 AM

Like okay, I can rename/remove some classes but I'm not 100% sure I want to get down deep into why the headers and columns don't line up well in this case.

Anbranin: Oh it looks weird because the headers are also centered. I did *not* see that.

Anbranin: I got to delete some stuff!

app/views/passengers/index.haml
```diff
@@ -15,30 +15,29 @@
...
         - unless @filters.include? 'Active'
-          %th.yesno Active
-        %th.yesno Permanent
-        %th.yesno Expiration Date
-        %th
-        %th
+          %th Active
+        %th Permanent
+        %th Expiration Date
+        %th{data: {orderable: 'false'}}
```
werebus: I'm on the fence about tossing centering for the icons.  If we want to put it back, though, Bootstrap has `.text-center`.

app/views/passengers/index.haml
```diff
@@ -46,9 +46,9 @@
           %td= passenger.email
           %td= passenger.phone
           - unless @filters.include? 'Active'
-            %td.yes= yes_no_image passenger.active?
-          %td.yes= yes_no_image passenger.permanent?
-          %td.yes{data: {sort: sortable_date(passenger.doctors_note)}}=passenger.expiration_display
```
werebus: Whoops. You _do_ want to keep the `data-sort` attribute, though.

app/views/passengers/index.haml
```diff
@@ -47,9 +46,9 @@
           %td= passenger.email
           %td= passenger.phone
           - unless @filters.include? 'Active'
-            %td.yes= yes_no_image passenger.active?
-          %td.yes= yes_no_image passenger.permanent?
-          %td.yes= passenger.expiration_display
+            %td= yes_no_image passenger.active?
```
werebus: Lastly, sorting these columns doesn't work because, as far as DataTables is concerned, they are empty cells. These also need a 'data-sort' attribute of some sort.

app/views/passengers/index.haml
Anbranin: I think I'm almost in favor of having it, just wanted to be minimalist. I'll put it back with bootstrap.

app/views/passengers/index.haml
Anbranin: woops, `dt=` was pressed.

app/views/passengers/index.haml
Anbranin: Okay so I want you to know that I know how it looks but DataTables really doesn't like straight up booleans and it needs strings.

werebus: I think it's actually a result of HAML's handling of booleans. It interprets the boolean to mean "should this data attribute be present in the DOM?" :

```
%div{data: {thing: true}}   -makes -  <div data-thing="">
%div{date: {thing: false}}  -makes-  <div>
```

Which is useful in other situations in this, because you can use `hasAttribute('data-thing')` to easily get your `true`s and `false`s into your Javascript logic.

app/assets/javascripts/datatables.js
```diff
@@ -1,32 +1,6 @@
...
-      { "bSortable": false },
-      null,
-      { "bSortable": false },
-      { "bSortable": false },
-    ],
-    "aLengthMenu": [[25, 50, 75, 100, -1], [25, 50, 75, 100, "All"]],
-    "iDisplayLength": 100
+  $('#passengers').DataTable({
+    destroy: true,
+    lengthMenu: [ [10, 25, 50, -1], [10, 25, 50, 'All'] ]
```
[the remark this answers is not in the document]
Anbranin: *pared

❤️

PR #127, opened 2019-09-30
Add archival

### LINKED ISSUE #120: Add status to passengers "Archived"
Passengers are marked inactive (active: false) when the expiration date for the doctor's note has passed. They're colored grey on the index page the moment the date passes, and then actually marked inactive in a `whenever` task every night and then can only be seen by clicking the "show inactive" filter on the index page.

However, it is now desired by the STSPVs that passengers actually have three states: "Active" - they need rides currently.
"Inactive" - their doctor's note has expired, but it could be an oversight and they could still need rides. After all, people get extensions a lot. They could easily be reactivated and should remain on the main index page, just a different color.
"Archived" - The passenger may come back at some point if they ever need our services again, but they probably (hopefully!) won't. Archived is a more "for real inactive" state, describing the differences between "the doctor's note just expired and they could legitimately still need our services and we should check with them" and "they healed up or moved away and don't need rides any more indefinitely."

The plan is that the STSPVs will manually move people to an archived state on whatever kind of regular basis they wish. Thus, the whenever task can be removed.

Concrete things that need to happen:

- [x] a new nav-tab on top that shows archived passengers on their own page
- [x] remove the "inactive" checkmark on the index page
- [x] remove the checkbox "show inactive" filter on the index page
- [x] remove the whenever task that makes passengers inactive
- [x] make an "archive" button - and conversely, a "de-archive" button.

Brainstorming:

I suspect in the end that there will be 3 or 4 statuses for passengers, now that we're implementing "archived" and will also soon implement "pending" - this is because passengers will fill out their own registration forms soon. I believe it makes sense to have "status" be an enum. So, "inactive" may not be necessary, because "inactive" should now be defined by someone having a doctor's note with an expiration date before today, and the only place that's necessary to display is on the index page. They'll be grey and that can be computed with a model method `#inactive`.

HOPE THIS ALL MAKES SENSE

Anbranin:
closes #120


app/controllers/passengers_controller.rb
```diff
@@ -1,8 +1,20 @@
 # frozen_string_literal: true
 
 class PassengersController < ApplicationController
-  before_action :find_passenger, only: %i[show edit update destroy]
-  before_action :access_control, only: %i[destroy]
+  before_action :find_passenger, only: %i[show edit update destroy toggle_archive]
+  before_action :access_control, only: %i[destroy archived toggle_archive]
+
+  def archived
```
Chauce: I like this much better than the previous "inactive" designation.

app/models/passenger.rb
```diff
@@ -34,13 +34,4 @@ def expiration_display
...
   end
-
-  def self.deactivate_expired_doc_note
-    expired = active.joins(:doctors_note)
-                    .where('doctors_notes.expiration_date < ?',
-                           DoctorsNote.grace_period)
-    expired.each do |passenger|
-      passenger.update_attributes active: false
-    end
-  end
```
Chauce: Are we no longer letting the database decide who is active based on doctor's notes?

Anbranin: There's no reason for a database attribute, no. The "inactive" people are highlighted on the index page just like before. I didn't see any reason to have a database attribute, although if you can see one I'm all ears.

PR #128, opened 2019-09-30
It's all about the restoration visits, baby.

### LINKED ISSUE #71: Navigating "back" to passengers index creates duplicate data table headers and footers

Anbranin:
closes #71 

This was an obscure one. I got the fix from [here](https://medium.com/philly-dev-shop/turbolinks-5-and-datatables-a882c29d6eff#.iksqjvhvb). The author of this article did state on [this informative issue discussion](https://github.com/turbolinks/turbolinks/issues/106) that it caused him some (unnamed) javascript issues down the line related to clicking the back button, but I haven't encountered any thus far.

Once the add-archival PR is merged, this'll be GTG

### DIFF
--- a/app/views/passengers/archived.html.haml
+++ b/app/views/passengers/archived.html.haml
@@ -1,3 +1,6 @@
+- content_for :head do
+  %meta{name: 'turbolinks-cache-control', content: 'no-cache'}
+
 - if @passengers.present?
   %h1 Archived Passengers
   %table#passengers
--- a/app/views/passengers/archived.html.haml
+++ b/app/views/passengers/archived.html.haml
@@ -1,6 +1,3 @@
-- content_for :head do
-  %meta{name: 'turbolinks-cache-control', content: 'no-cache'}
-
 - if @passengers.present?
   %h1 Archived Passengers
   %table#passengers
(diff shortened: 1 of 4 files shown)


werebus:
I guess my question is: the gist of that blog post is, "if you're using AJAX-based DataTables, there's no point in caching."  But... we aren't using AJAX-based DataTables. It seems like it would be better to pursue the "un-datatable it before caching" solution:

```javascript
$( document ).on('turbolinks:load', function(){
  $('#passengers').DataTable({
    //remove destroy: true
    stateSave: true,
    lengthMenu: [ [10, 25, 50, -1], [10, 25, 50, 'All'] ]
  });
});

$( document ).on('turbolinks:before-cache', function() {
  $('#passengers').DataTable().destroy();
});
```

since all of the _data_ in the table is sent at once and a cache of that data is likely valid and a performance boost.

app/views/passengers/archived.html.haml
```diff
@@ -1,3 +1,6 @@
+- content_for :head do
+  %meta{name: 'turbolinks-cache-control', content: 'no-cache'}
+
```
Chauce: So simple ... 👍

PR #132, opened 2019-10-04
Check existing pax

### LINKED ISSUE #121: Search for existing passengers when making new ones - start tracking spire IDs
Related to #120 
In the current implementation of this database (I think it's a google spreadsheet, or maybe they're still using the access database) there are multiple passengers with the same information because once a passenger is inactive, they disappear from view--and searching for them is harder than just creating a new passenger. Lots of duplicate data.
The idea now is to start storing Spire IDs (at least of new passengers, existing ones fitting into the new schema are a problem someone will have to tackle, or maybe not) and when a new passenger is being created, our code will search the existing passengers (archived or not) and if a passenger already exists with that Spire ID, a friendly message should pop up and redirect the user to the page for that passenger.
It would be best if this happened not upon submit when the user has already typed in all sorts of information, but with an Ajax call triggered by filling in (and de-focusing) the Spire ID field.
A user could also search the "archived" page for this passenger, but implementing a fancy search function is an enhancement to think about, and command-F also works.

Anbranin:
closes #121 

Can't write tests yet, need system tests for the javascript support.


app/assets/javascripts/passengers.js
```diff
@@ -0,0 +1,15 @@
+$( document ).on("turbolinks:load", function() {
+  $('#passenger_spire').focusout(function(){
+    $.ajax({
+      type: 'GET',
+      url: '/passengers/check_existing',
+      data: { spire_id: $(this).val() },
+      success: function(response) {
+        if(response == undefined){ return }
+        $('body').append('<div id="check-existing" class="modal" role="dialog" tabindex= "-1">')
```
werebus: So, why not have this tag in the HAML?

app/assets/javascripts/passengers.js
```diff
@@ -0,0 +1,15 @@
...
+        $('#check-existing').html(response)
+        $('#check-existing').modal()
```
werebus: I _think_ you can chain these together. That is, `$('#check-existing').html(response).modal()`

app/views/passengers/_check_existing.haml
```diff
@@ -0,0 +1,9 @@
+.modal-dialog{role: 'document'}
+  .modal-content
+    .modal-header
+      %h5 A passenger already exists for this spire ID. Go to edit that passenger?
+      %button.close{data: {dismiss: 'modal', 'aria-label':'Close'}}
+        %span{'aria-hidden': 'true'} &times;
+    .modal-footer
+      %button.btn.btn-default{data: {dismiss: 'modal', 'aria-label':'Close'}} Keep adding new
+      = link_to 'Edit existing passenger', edit_passenger_url(@passenger), class: 'btn btn-primary'
```
werebus: Some verb agreement / parallel construction nit pick. Perhaps:

*  A passenger already exists for this Spire ID. Edit that passenger? (Spire should be capitalized)
*  Add new passenger
*  Edit existing passenger

app/views/passengers/_check_existing.haml
```diff
@@ -0,0 +1,9 @@
+.modal-dialog{role: 'document'}
+  .modal-content
+    .modal-header
+      %h5 A passenger already exists for this spire ID. Go to edit that passenger?
+      %button.close{data: {dismiss: 'modal', 'aria-label':'Close'}}
+        %span{'aria-hidden': 'true'} &times;
+    .modal-footer
+      %button.btn.btn-default{data: {dismiss: 'modal', 'aria-label':'Close'}} Keep adding new
```
werebus: What does `.btn-default` do?

app/assets/javascripts/passengers.js
```diff
@@ -0,0 +1,15 @@
+$( document ).on("turbolinks:load", function() {
+  $('#passenger_spire').focusout(function(){
+    $.ajax({
+      type: 'GET',
+      url: '/passengers/check_existing',
+      data: { spire_id: $(this).val() },
+      success: function(response) {
+        if(response == undefined){ return }
```
werebus: This seems odd. Why would the response be `undefined`?

app/controllers/passengers_controller.rb
```diff
@@ -16,6 +16,11 @@ def toggle_archive
     redirect_to passengers_url, notice: 'Passenger successfully updated'
   end
 
+  def check_existing
+    @passenger = User.find_by(spire: params[:spire_id])
+    render partial: 'check_existing'
```
werebus: Doesn't this controller action _always_ render this partial?  Maybe I'm missing something?

app/assets/javascripts/passengers.js
Anbranin: It's `undefined` when the controller responds with nothing. Open to suggestions to make it less weird.

app/controllers/passengers_controller.rb
Anbranin: yes and it finds a user not a passenger

app/assets/javascripts/passengers.js
werebus: Eh. Maybe my objection is just about naming?

The "response" is there, it's the response _body_ (or whatever) that's blank.

app/views/passengers/_form.html.haml
```diff
@@ -42,8 +42,8 @@
      options_from_collection_for_select(MobilityDevice.order(:name),
       :id, :name), include_blank: 'None'
   .field
-    = form.label :active
-    = form.check_box :active, disabled: !@current_user.admin?
+    = form.label :archived?
+    = form.check_box :archived?
```
werebus: Did you know you can specify what the checked and unchecked values of a checkbox are?

```haml
= form.check_box :archived?, {}, 'archived', 'active'
```

Then that value will be in the `params` and you can use it directly.

app/views/passengers/_check_existing.haml
Anbranin: Why I chose it: It had the color I wanted and it seemed to fit lexically as well since it's a "dismissal" like, no, stay with the default action which is to just keep going. 
As for what it "does" it just provides styling, that's it.

werebus: Yeah. My question came from the fact that I didn't know we were still on Bootstrap 3 on this app.  Bootstrap 4 doesn't have this class.

werebus:
This PR should probably address #113 

But otherwise just my wording nit-picks left.

Anbranin:
> This PR should probably address #138

I just tried to fix it and realized it wasn't broken. I did hurredly resolve a bunch of merge conflicts this morning in my attempt to get a "demo" branch going, so I blame that.

I just have to make sure when I do actually merge everything it doesn't break it.

PR #134, opened 2019-10-07
Add doctors note checkbox

### LINKED ISSUE #118: Add checkbox for "has doctor's note" and conditional validation on expiration date
Right now an expiration date existing is how we determine whether a passenger has a doctor's note. It's really the only thing about a doctor's note.
But, the supervisors would like a way–to more easily than that–visually determine who has a doctor's note (especially now that the expiration date could be filled in with the date 3 days from registration).
I envision this: The checkbox "has doctor's note" is on the passenger show/edit form. "I exist" is a property on DoctorsNote (let me explain...), so that we can easily conditionally validate that the expiration date exists if the "has doctor's note" box is checked on the passenger page.

### LINKED ISSUE #119: Add "registration date" as a field to the passenger show page
Registration date is just passenger#created at but we should show people.

Anbranin:
closes #118 
closes #119 also because I rather needed the registration date field to add the functionality for an alternative expiration display.


app/views/passengers/show.haml
```diff
@@ -31,11 +31,20 @@
   %td.yes= yes_no_image @passenger.permanent
 - if @passenger.temporary?
   %p
-    %b Doctor's Note Expiration Date:
-    = @passenger.expiration_display
+    %b Has Doctor's Note?
+    %td.yes= yes_no_image @passenger.doctors_note.present?
```
jforgue: Image? why?

spec/models/passenger_spec.rb
```diff
@@ -35,4 +35,31 @@
...
+  describe 'rides_expire' do
+    context 'permaent passenger' do
+      it 'returns nil' do
+        @passenger.update permanent: true
+        expect(@passenger.rides_expire).to be nil
+      end
+    end
+    context 'doctors note present' do
+      it 'returns the expiration date of the note' do
+        date = 14.days.since.to_date
```
jforgue: Similarly, `Date.today + 14`

app/assets/javascripts/passengers.js
```diff
@@ -1,4 +1,12 @@
 $( document ).on("turbolinks:load", function() {
+  $('#passenger_has_doctors_note').on('change', function(){
+    $('#passenger_doctors_note_attributes_expiration_date').prop('required', $(this).prop('checked'));
```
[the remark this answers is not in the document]
jforgue: Whoops, nevermind.

app/views/passengers/show.haml
Anbranin: I didn't name that method. Also, it's an image, so.

app/models/passenger.rb
```diff
@@ -24,12 +25,29 @@ class Passenger < ApplicationRecord
...
     doctors_note.try(:expiration_date).try :strftime, '%m/%d/%Y' || 'No Note'
   end
 
+  def rides_expire
+    return if permanent?
+
+    return doctors_note.expiration_date if doctors_note.present?
+    return registration_date + 3.days if persisted?
+
+    3.days.since.to_date
```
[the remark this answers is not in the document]
Anbranin: I like using the method `since` much more than the plus sign, personally.

spec/models/passenger_spec.rb
Anbranin: Again I respectfully disagree that the plus sign is clearer. To me, it makes me look for math first. Although perhaps a discussion about why you think it's unclear is warranted.

app/models/passenger.rb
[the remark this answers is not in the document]
werebus: Personally, I've never been able to parse how the word "since" on its own means a date or time in the future. Luckily, Rails agrees with me and has an alias:

```ruby
4.days.from_now
```

`.since` reads much better with the optional argument:

```ruby
4.days.since(my_birthday)
4.days.after(my_birthday)
```

app/views/passengers/show.haml
werebus: :neckbeard: Well, actually, it's a font glyph.

Anbranin: Which is a pictoral representation of a character. Everything we look at is technically an image. Fight me

app/models/passenger.rb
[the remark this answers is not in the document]
Anbranin: I will choose `from_now` to resolve this point of contention since I agree that `since` isn't a great method name (although still better than a plus)

app/views/passengers/show.haml
Anbranin: So changing the method name could be done, it's not a great method name, but if we go down that rabbit hole with this PR it'll never end.

app/assets/stylesheets/passengers.scss
```diff
@@ -115,6 +115,6 @@ div {
   }
 }
 
-.hide_view {
+.hide-view {
```
werebus: FWIW, Bootstrap has `.hidden` already.

app/models/passenger.rb
```diff
@@ -24,12 +25,29 @@ class Passenger < ApplicationRecord
...
     return if permanent?
 
     doctors_note.try(:expiration_date).try :strftime, '%m/%d/%Y' || 'No Note'
   end
 
+  def rides_expire
+    return if permanent?
+
+    return doctors_note.expiration_date if doctors_note.present?
+    return registration_date + 3.days if persisted?
```
werebus: `3.days.since(registration_date)`?  :stuck_out_tongue_winking_eye:

app/views/passengers/show.haml
werebus: Why is this a `td`? I don't see a `table`. Seems like invalid HTML

db/migrate/20191007174525_add_registration_date_to_existing_passengers.rb
```diff
@@ -0,0 +1,5 @@
+class AddRegistrationDateToExistingPassengers < ActiveRecord::Migration[5.2]
+  def change
```
werebus: Seems like this should be `def up`

app/views/passengers/index.haml
```diff
@@ -43,7 +43,7 @@
           %td= passenger.email
           %td= passenger.phone
           %td.text-center{data: {sort: passenger.permanent?.to_s}}= yes_no_image passenger.permanent?
-          %td.text-center{data: {sort: sortable_date(passenger.doctors_note)}}=passenger.expiration_display
+          %td.text-center{data: {sort: sortable_date(passenger.doctors_note)}}=passenger.rides_expire.try(:strftime, '%m/%d/%Y')
```
werebus: Do you want to make this change to the "archived" view too?  How about the printout?

app/assets/stylesheets/passengers.scss
Anbranin: I was looking for that!!

app/models/passenger.rb
Anbranin: LOL

app/assets/stylesheets/passengers.scss
Anbranin: OK, I tried this out--doing this makes Jquery's `toggle()` not work, because of the `!important` aspect of the `display: none` property of the lass `.hidden`, which, you pointed out, it still somehow more important than inline styles.

app/views/passengers/index.haml
Anbranin: It was a conscious decision not to make this change to the archived view. Since these passengers aren't actively getting rides, I couldn't find a reason why it might be useful, as opposed to knowing when someone's doctor's note expired, which might be.

As for the printout, I plumb forgot.

app/views/passengers/show.haml
jforgue: I guess what I really meant to ask was, what does it look like? Why is that better than basic text? But I digress.

spec/models/passenger_spec.rb
Anbranin: I'll change this to `from_now`

jforgue: There are two reasons I suggest the `+`:

1. `.since` is kind of confusing
2. using `14.days.since.to_date` you have to go from a `DateTime` to a `Date` for no real reason other than it looks subjectively better

Although looking at Matt's comment I agree that `.from_now` is much more readable.

PR #148, opened 2019-10-11
Longer rides

### LINKED ISSUE #116: Add way to visualize "needs longer ride"
Right now the mobility device management interface will show you which kind of mobility device needs a longer ride (it's an attribute on the object), but there's no way to easily see that from the passenger index.
It was requested that there be a way to see at-a-glance, specifically with styling and _not_ with another column, which passengers need more than a single ride slot.

Chauce:
We have stripes!
Closes #116 

Screen Shot 2019-10-11 at 1 36 29 PM

### DIFF
--- a/app/assets/stylesheets/passengers.scss
+++ b/app/assets/stylesheets/passengers.scss
@@ -36,6 +36,16 @@ table#key, table#passengers {
   .no_note {
     background-color: #90C3D4;
   }
+  .longer_ride {
+    /* font-weight: bold; */
+    background: repeating-linear-gradient(
+    45deg,
+    rgba(0, 0, 0, 0),
+    rgba(0, 0, 0, 0) 10px,
+    rgba(0, 0, 0, 0.1) 10px,
+    rgba(0, 0, 0, 0.1) 20px
+  )
+  }
 }
 
 .passenger-filters{
--- a/app/assets/stylesheets/passengers.scss
+++ b/app/assets/stylesheets/passengers.scss
@@ -38,7 +38,7 @@ table#key, table#passengers {
   }
   .longer_ride {
     /* font-weight: bold; */
-    background: repeating-linear-gradient(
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/passengers.scss
```diff
@@ -36,6 +36,16 @@ table#key, table#passengers {
   .no_note {
     background-color: #90C3D4;
   }
+  .longer_ride {
+    /* font-weight: bold; */
```
werebus: Don't need this comment

app/helpers/passengers_helper.rb
```diff
@@ -11,13 +13,30 @@ def doctors_note_row_class(note)
     end
   end
 
-  def passengers_table_row_class(passenger)
-    return if passenger.permanent?
-    return 'no_note' if passenger.doctors_note.blank?
+  def longer_ride(device)
```
werebus: Probably good to stick with the convention that these class helpers end with `_class`

Chauce:
Code climate doesn't like the complexity ... I am open to options for paring down the logic.

app/helpers/passengers_helper.rb
```diff
@@ -11,13 +13,30 @@ def doctors_note_row_class(note)
...
+    if device.try(:needs_longer_rides) == true
+      'longer_ride'
+    end
+  end
 
-    doctors_note_row_class(passenger.doctors_note)
+  def passengers_table_row_class(passenger)
+    if passenger.permanent? && longer_ride(passenger.mobility_device).nil?
```
werebus: This method seems like it has too many conditionals.

What you're interested in doing is joining two `String`-or-`nil`s together, right?  Actually, you don't even have to join them together. The way this method is used, you could also return an array of strings.

Anbranin:
We decided against doing stripes.

PR #151, opened 2019-10-17
Add fixed headers and improved styling

### LINKED ISSUE #60: Better sort order arrows
I'm sorting by Name, however, it looks like I'm sorting by Mobility Device instead. I would prefer better placement for those arrows so people don't get confused when using the sort functions.
screen shot 2017-08-31 at 12 34 25 pm

Anbranin:
For styling:
- datatables.net-bs/css/dataTables.bootstrap.css styles the datatables nicely, giving you big ole sort arrows that closes #60 
- datatables.net-fixedheader-bs/css/fixedHeader.bootstrap.css allows
the fixedHeader plugin for datatables to work
~datatables.net-dt/css/jquery.dataTables.css styles datatables such
that the datatable will take up the entire width of the page instead of
being squished. It has more padding in between cells as well. It just
looks nicer, except for the arrows that the bootstrap css files get you.~ is unneccessary.

For javascript:
- datatables.net/js/jquery.dataTables.js allows you to initialize
datatables like at all.
- datatables.net-fixedheader/js/dataTables.fixedHeader.js is the meat of
the fixedHeader plugin and, you kow, fixes the header.
- datatables.net-bs/js/dataTables.bootstrap.js is the least necessary of
this bunch but does give the search bar a class which allows nice
form-control bootstrap styling. It looks better so I'm keeping it.

shots:
Screen Shot 2019-10-17 at 12 33 13 PM
Screen Shot 2019-10-17 at 12 33 06 PM


app/assets/stylesheets/application.scss
```diff
@@ -10,6 +10,9 @@
  * files in this directory. Styles in this file should be added after the last require_* statement.
  * It is generally better to create a new file per style scope.
  *
+ *= require datatables.net-bs/css/dataTables.bootstrap.css
+ *= require datatables.net-fixedheader-bs/css/fixedHeader.bootstrap.css
+ *= require datatables.net-dt/css/jquery.dataTables.css
```
werebus: Datatables docs say don't do both.

>  For CSS only a single file should be included as the styling library will provide the core styling options for the table. The CSS files have the format dataTables.{style}.css (dataTables.bootstrap.css for Bootstrap for example).

As a hint:  Try adding the class `.table` to the table to get it to be a [Bootstrap table](https://getbootstrap.com/docs/3.3/css/#tables)

>  For basic styling — light padding and only horizontal dividers — add the base class `.table` to any `<table>`. It may seem super redundant, but given the widespread use of tables for other plugins like calendars and date pickers, we've opted to isolate our custom table styles.

package.json
```diff
@@ -2,8 +2,9 @@
   "name": "st_pax_tracker",
   "private": true,
   "dependencies": {
-    "datatables.net": "^1.10.15",
```
werebus: I think it would be a good idea to keep this dependency explicit

Anbranin: Since I can't edit that file directly, how exactly does that work? Do I yarn add (thing-at-specific-version-number)?

werebus:
Arguable that this also closes #60

Anbranin:
yes. I believe these arrows are less confusing and resolves that issue.

package.json
werebus: >  ... I can't edit that file directly ...

Oh?

Anbranin: I thought it was like a Gemfile.lock! It's not?

Anbranin: Like I definitely _didn't_ edit it directly this time, let's just put it that way.

werebus: `Gemfile` : `package.json` :: `Gemfile.lock` : `yarn.lock`

Anbranin: Oooh.

app/assets/stylesheets/application.scss
Anbranin: Sure but that takes away the arrows.

Anbranin: I guess I can leave the original arrows and it won't close #60 and that's fine.

Anbranin: I am wrong! I removed the wrong one. We're good.

Anbranin:
> I'd also recommend removing the size-fiddling

I had no idea that was happening and the CSS rules in place were driving me crazy. Thank you.

package.json
```diff
@@ -3,7 +3,9 @@
   "private": true,
   "dependencies": {
     "datatables.net": "^1.10.15",
-    "datatables.net-dt": "^1.10.15",
+    "datatables.net-bs": "^1.10.20",
+    "datatables.net-dt": "^1.10.20",
```
werebus: So, the question is, do you need the `-dt` package now?

werebus: (You aren't including anything from it)

Anbranin: Oh gosh right

werebus:
Just out of curiosity, does the "archived" view still look acceptable?

Anbranin:
It does not. I 100% forgot about that one, thanks.

Issue #88, opened 2017-09-25
Test coverage is down


jforgue:
Test coverage is sitting comfortably around 97%. I don't think it'll be too challenging to get the app fully tested.

werebus:
So, it turns out that SimpleCov wasn't configured correctly. We're really at about 45%

jforgue:
This issue probably isn't necessary. We know what we are.

PR #157, opened 2019-10-24
SimpleCov wasn't configured for Rails

werebus:
Which meant that the coverage number included everything in `config/` and `spec/`, all of which was run by the specs.

Also took some time to tweek the RSec config in general.

Addresses #88 (in that it reveals that coverage is actually not very good)


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 97.17
+    "covered_percent": 45.58
```
werebus: :tada:

spec/spec_helper.rb
```diff
@@ -14,20 +17,20 @@
   config.mock_with :rspec do |mocks|
     mocks.verify_partial_doubles = true
   end
-  config.shared_context_metadata_behavior = :apply_to_host_groups
-end
 
-def when_current_user_is(user)
```
werebus: Moved to a `spec/support` file.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 97.3
+    "covered_percent": 47.1
```
Anbranin: wtf LOL

spec/helpers/passengers_helper_spec.rb
```diff
@@ -12,7 +12,7 @@
 #     end
 #   end
 # end
-describe PassengersHelper do
+RSpec.describe PassengersHelper do
```
Anbranin: Is this the monkey patching?

coverage/.last_run.json
werebus: :tada:

spec/spec_helper.rb
```diff
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
+require 'pathname'
 require 'simplecov'
 
-SimpleCov.start do
+Pathname(__dir__).join('support').glob('**/*.rb').each { |f| require f }
+
+SimpleCov.start 'rails' do
```
werebus: This is the real "fix".

spec/helpers/passengers_helper_spec.rb
werebus: Yes. RSpec monkey-patches `Object` to include a `.describe` method. But they kinda regret it.

Anbranin: Meanwhile I'm over here in my other apps *coughumaps-railscough* removing every `Rspec` I see because I'm not used to seeing it and it seems wrong

PR #159, opened 2019-10-24
Skip validations

### LINKED ISSUE #153: Don't validate when archiving
A lot of invalid passengers exist in the database due to the presence of spire ID being validated. As it has been many months since it was used many passengers must be archived. The ST Supervisors wish to archive without having to enter in more information about the passenger, which is reasonable.

Anbranin:
closes #153

### DIFF
--- a/app/models/passenger.rb
+++ b/app/models/passenger.rb
@@ -25,17 +25,7 @@ class Passenger < ApplicationRecord
 
   belongs_to :mobility_device, optional: true
 
-  before_validation do
-    assign_registration_date
-  end
-
-  def assign_registration_date
-    if active_status_changed? && active?
-      assign_attributes(registration_date: Time.zone.today)
-    elsif registration_date.blank?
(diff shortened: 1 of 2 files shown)


app/models/passenger.rb
```diff
@@ -26,6 +26,10 @@ class Passenger < ApplicationRecord
   belongs_to :mobility_device, optional: true
 
   before_validation do
+    assign_registration_date
```
werebus: If you're moving this to a method, you can refer to it by symbol:

```ruby
before_validation :assign_registration_date
```

app/models/passenger.rb
```diff
@@ -26,6 +26,10 @@ class Passenger < ApplicationRecord
...
+  end
+
+  def assign_registration_date
```
werebus: My instinct is that this should be a private method. Do you anticipate using it elsewhere?

Anbranin: ha no, there just weren't any existing private methods so I forgot they were a thing.

PR #162, opened 2019-10-28
add exception notifier gem and configure mail settings

### LINKED ISSUE #161: Add exception notifier gem
It was crazy trying to debug the application without exception notifier. What an oversight!

Anbranin:
closes #161

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -69,12 +69,12 @@ Rails.application.configure do
   # environment)
   # config.active_job.queue_adapter     = :resque
   # config.active_job.queue_name_prefix = "st_pax_tracker_#{Rails.env}"
-  config.action_mailer.perform_caching = false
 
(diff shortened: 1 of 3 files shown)


config/environments/production.rb
```diff
@@ -100,4 +100,10 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  config.middleware.use ExceptionNotification::Rack,
+    email: {
+      sender_address: %{'st-pax' <transit-it@admin.umass.edu>},
```
LiamBrandt: Is this missing a `w` for the `%w`? Also, should we be using `%w{}` instead of `%w[]`?

Anbranin: Nope, it's that way in round-three and the exception_notifier docs. It resolves to a string, not an array of strings.

Anbranin: https://github.com/smartinez87/exception_notification

PR #168, opened 2019-11-01
Passenger registration

Anbranin:
People can look this over and judge me as we go


app/views/passengers/index.haml
```diff
@@ -18,10 +18,9 @@
   %table#key
     %tbody
       %tr
-        %td.will_expire_soon Will Expire Within 1 Week
-        %td.expired_within_grace_period Expired Within Grace Period
-        %td.inactive Inactive
-        %td.no_note No Doctor's Note
+        %td.expires-soon Expires within one week
+        %td.needs-note Within grace period, needs doctor's note
```
avacmeyer: This is definitely more clear.

Anbranin:
So, this PR and master have diverged too much for me to bring them back together easily--we ended up going in different directions than I had imagined. I'm going to use this as a reference because I think there are some good ideas in here even though the code isn't usable, and after that will be deleting the branch.

PR #166, opened 2019-10-30
Flash

Anbranin:
Instead of displaying errors per view with some view logic, I used the flash implementation seen in most of our other apps. The class is given by the controller along with the layout and allows us to use [bootstrap alert styling](https://getbootstrap.com/docs/4.0/components/alerts/).

PR #170, opened 2019-11-05
Flash changes for feedback

werebus:


werebus:
This is a PR to add onto #166

PR #171, opened 2019-11-06
Restore datatables bootstrap theme

werebus:
It was there in #151, which was a PR against #150. But, when #150 was merged, it was gone.


Anbranin:
:-o

PR #172, opened 2019-11-06
Move the view filtering to DataTables

werebus:
DataTables can filter. 

Advantages:

1. Much faster
2. No more confusion about what the print button will print: The view is updated the moment you change the filter, so the button will print which radio is selected, but that is the same as what is on the screen.


app/assets/javascripts/datatables.js
```diff
@@ -1,8 +1,9 @@
 $( document ).on("turbolinks:load", function() {
   $('#passengers').DataTable({
     fixedHeader: true,
+    lengthMenu: [ [10, 25, 50, -1], [10, 25, 50, 'All'] ],
+    order: [[0, 'asc']],
```
werebus: This isn't strictly related, but if the view starts out ordered (it does), the UI should reflect that.

app/controllers/passengers_controller.rb
```diff
@@ -79,18 +71,13 @@ def destroy
 
   private
 
-  def passenger_params
```
werebus: Sorry, I reordered the private methods.

app/controllers/passengers_controller.rb
```diff
@@ -102,13 +89,28 @@ def disallow_nonexpiring_note(permitted_params)
...
+    @passenger = Passenger.find(params[:id])
+  end
+
+  def passenger_params
+    base_passenger_params
+      .then { |p| restrict_admin p }
+      .then { |p| disallow_nonexpiring_note p }
+  end
+
+  def passenger_pdf
```
werebus: This is the only private method that was added or changed.

app/controllers/passengers_controller.rb
```diff
@@ -102,13 +89,28 @@ def disallow_nonexpiring_note(permitted_params)
...
+    @passengers = @passengers.send(@filter) if @filter.present?
+    pdf = PassengersPDF.new(@passengers, @filter)
+    name = "#{@filter} Passengers #{Date.today}".capitalize
```
Anbranin: So if filter isn't present the name will just be " Passengers etc"? Is it purposeful that you're doing away with "All Passengers" as a possible title?
A possible elegant solution if you did want to keep that at all would be to have `@filter ||= 'all'` in the beginning of the method.

app/assets/javascripts/passengers/filtering.js
```diff
@@ -1,6 +1,13 @@
 function updatePassengerFilter() {
-  var selected = $("input[name=filter]:checked").val();
+  if (!$(".passenger-filters input").length) return;
+
+  var selected = $(".passenger-filters input[name=filter]:checked").val();
   var column = $("table#passengers th").index($(".permanent"));
+
+  $("span.filter-name").text(
+    selected.charAt(0).toUpperCase() + selected.slice(1)
```
Anbranin: Ok Javascript.

PR #174, opened 2019-11-07
Xpath removal

werebus:
These tests failed when I decided destroying a user should be a button rather than a link.

In fixing it, I realized that the xpath notation is nearly impenetrable to me, and it doesn't really reflect how a "real person" would find what thing to click on. Instead of finding by `href`, let's find the "Destroy" that's in the same row as the user's name.

### DIFF
--- a/spec/system/user_management_spec.rb
+++ b/spec/system/user_management_spec.rb
@@ -13,10 +13,10 @@ RSpec.describe 'User Management', js: true do
     end
     it 'puts the errors in the flash' do
       visit users_url
-      # Find destroy link by href because there are 2 users by necessity.
-      # Although you definitely _can_ delete yourself.
-      find(:xpath, "//a[@href='/users/#{@user.id}']").click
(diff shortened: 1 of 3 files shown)


spec/system/user_management_spec.rb
```diff
@@ -13,10 +13,10 @@
     end
     it 'puts the errors in the flash' do
       visit users_url
-      # Find destroy link by href because there are 2 users by necessity.
-      # Although you definitely _can_ delete yourself.
-      find(:xpath, "//a[@href='/users/#{@user.id}']").click
-      page.driver.browser.switch_to.alert.accept
+      # Click the 'destroy' link in the user's row
+      page.accept_confirm 'Are you sure?' do
+        within('tr', text: @user.name) { click_on 'Destroy' }
```
Anbranin: It was this easy? I knew there must be a way to do it but I admit, I was too lazy to look it up at the time and thought it must be more impenetrable than finding the button by xpath.

PR #186, opened 2019-11-15
Add verification source

Anbranin:
We need the existence of two kinds of objects in the database, as @werebus put it: "Who, other than you, thinks you need van rides?" and "what do they say?"

"What do they say" is "eligibility verification". "Who are they" is "verifying agency." I can do this in two steps and will open another PR that renames "doctor's note" (which isn't an apt name any more) to EligibilityVerification.

I've replaced "registered with disability services?" with "verifying agency" here, and made it so permanent passengers can also be "verified" (although they may not have an expiration date).


Anbranin:
Decided on a different naming convention.

PR #194, opened 2019-11-25
Passenger PDF Layout changes

### LINKED ISSUE #176: Change printout to letter paper
Because we try to make it that way:

https://github.com/umts/st-pax/blob/526add8ad27d2a5af2de81be00300f9dc783ca4c/app/pdfs/passengers_pdf.rb#L8

If, as I suspect, they usually go on landscape _letter_ paper, that should be changed because the two papers don't share an aspect ratio.

Tanvish:
- Closes #176 
- Removed the tabloid layout from the PDF and added a header to the table…s after the start of every new page.

### DIFF
--- a/app/pdfs/passengers_pdf.rb
+++ b/app/pdfs/passengers_pdf.rb
@@ -4,7 +4,7 @@ require 'prawn/table'
 
 class PassengersPDF < Prawn::Document
   def initialize(passengers, filter)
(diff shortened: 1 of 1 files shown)


Tanvish:
# Supplementary screenshot
Screen Shot 2019-11-25 at 2 54 22 PM

Anbranin:
Nice

PR #195, opened 2019-11-25
Rename doctors note

Anbranin:
As we move towards not accepting doctor's notes any more, the model name "DoctorsNote" doesn't sit as well with me.

There are no functional changes in this PR, only a name change, but it sets us up to have a more graceful movement to the new modus operandi: Storing which agency verifies that the passenger in question is eligible to receive rides and being able to contact them directly.


Anbranin:
⚡️ plz 😭

jforgue:
Looks good. I see a few instances of 'doctors note' left but I'm guessing those are specific to the field itself. But why is coverage dropping?

Anbranin:
I removed one model line. A coverage change is appropriate.

PR #201, opened 2019-12-19
A button added to copy all the emails of passengers displayed to clipboard.

### LINKED ISSUE #144: add "copy emails of filtered list to clipboard" button next to "print these filters" button
Admins regularly have to email passengers in bulk with announcements--they'd like a button to "copy all emails to clipboard" much like we have in umasstransit with the directory.

Tanvish:
- Closes #144 
Screen Shot 2019-12-19 at 10 24 14 AM


app/assets/javascripts/application.js
```diff
@@ -19,4 +19,5 @@
 //= require rails-ujs
 //= require turbolinks
 //= require moment/moment
-//= require_tree .
+//= require clipboard/dist/clipboard
+//= require_tree .
```
Anbranin: Looks like you removed the line at the end of the file, can you set your text editor to not do this?

Anbranin: (and remove that particular change)

package-lock.json
```diff
@@ -0,0 +1,105 @@
+{
```
Anbranin: Tried to leave a comment spanning the whole file but couldn't...

After reading this StackOverflow article it's clear to me how the `package-log.json` differs from `yarn.lock`. And it would be, perhaps, nicer for us to have both files in our repository since this is open source, like they suggest. However, we need to, like one user pointed out, have in that case an process that makes sure those files stay in sync and we don't.
I would therefore move a change of this sort to its own branch.
For now please use `yarn` to install your packages and make sure that `yarn.lock` gets updated, and re-run all your tests.

app/assets/javascripts/application.js
Tanvish: Aight, done.

package-lock.json
Tanvish: Aight, done.

app/assets/javascripts/passengers/passengers.js
```diff
@@ -7,6 +7,17 @@ $( document ).on("turbolinks:load", function() {
     $('.doctors-note-fields').toggle(!$(this).prop('checked'));
   });
 
+  new ClipboardJS('#copybtn', {
+    text: function() {
+      var arr = $("tbody tr")
```
werebus: Perhaps a selector that specifies the _passengers_ table?

app/views/passengers/index.haml
```diff
@@ -14,6 +14,9 @@
               checked: @filter == filter,
               inline: true
       = f.submit 'Print'
+      -if @current_user.admin?
+        .btn.btn-secondary.pull-right#copybtn Copy Emails
```
werebus: Seems like this button shouldn't be _within_ the "print" form.

app/assets/javascripts/passengers/passengers.js
werebus: I think you can do this in a few fewer steps:

```js
text: function() {
  $("tbody tr").map(function(){return $(this).data("email")}).get().join(";")
}
```

app/views/passengers/index.haml
werebus: Also, `pull-right` seems a little weird.

app/assets/javascripts/passengers/passengers.js
Tanvish: This doesn't seem to work though.

app/views/passengers/index.haml
Tanvish: Okay, I guess the button for copying can be below the button for print. Screenshot
 attached:

Screen Shot 2020-01-09 at 1 58 58 PM

app/assets/javascripts/passengers/passengers.js
werebus: Sorry. It's missing the `return`:

```js
    text: function() {
      return $("#passengers tbody tr").map(function(){
        return $(this).data("email")
      }).get().join(";")
```

app/assets/javascripts/passengers/passengers.js
```diff
@@ -1,4 +1,22 @@
...
+    $('#passenger_doctors_note_attributes_expiration_date').prop('required', $(this).prop('checked'));
+  });
+
+  $('#passenger_permanent').on('change', function(){
+    $('.doctors-note-fields').toggle(!$(this).prop('checked'));
+  });
+
+  new ClipboardJS('#copybtn', {
+    text: function() {
+      var arr = $("#passengers body tr")
```
werebus: Whoops.  that should be `tbody`, not `body`.

app/views/passengers/index.haml
```diff
@@ -14,6 +14,10 @@
               checked: @filter == filter,
               inline: true
       = f.submit 'Print'
+    -if @current_user.admin?
+      %br
```
werebus: No `<br>` tags, please.

Note that I kinda prefer the button to the right rather than below. I just don't know if it should be within the `form` DOM-wise.

Tanvish: I can't seem to figure out a way to do that. I might need some help.

app/assets/javascripts/passengers/passengers.js
```diff
@@ -1,4 +1,19 @@
...
+      return $("#passengers tbody tr").map(function(){
+        return $(this).data("email")
+      }).get().join(";")
+    }
+  });
+  clipboard.on('success', function() {
+    $('#copybtn').tooltip({title: "Copied"}).tooltip('show');
+  });
+
+  clipboard.on('error', function() {
```
Tanvish: I couldn't find a way to merge this into one block.

werebus: Yeah, all the ways I could think of would make Code Climate happier, but actual developers sadder. For instance:

```js
  var clipResults = {'success': 'Copied', 'error': 'Falied to copy!'}
  for (const event in clipResults) {
    if (!clipResults.hasOwnProperty(event)) continue;
    clipboard.on(event, function() {
      $('#copybtn').tooltip({title: clipResults[event]}).tooltip('show');
    });
  }
```

But it: a) Requires ES5+, b) is the same number of lines of code, and c) adds a level of inderection that makes the code _harder_ to understand.

Tanvish:
Just an FYI, I don't have the authorization to merge this PR.

PR #203, opened 2019-12-23
Add verifying agency

Anbranin:
#195 needs to be merged first. I'm also going to write some tests, but the meat of this is ready to go and the visible changes have been checked out with the ST folks.


app/models/eligibility_verification.rb
```diff
@@ -2,11 +2,24 @@
 
 class EligibilityVerification < ApplicationRecord
   belongs_to :passenger
+  # belongs_to set to optional true with presence validation in order to
+  # make bootstrap validations work nicely.
+  belongs_to :verifying_agency, optional: true
```
werebus: You might not want  `optional: true`. This allows for a `verifying_agency_id` that doesn't go with any actual `verifying_agency`.

app/models/eligibility_verification.rb
```diff
@@ -17,24 +30,38 @@ def self.expiration_warning
   end
 
   def will_expire_within_warning_period?
+    return if expiration_date.blank?
```
werebus: I know it's functionally the same, but I prefer predicate methods to always return `true` or `false`

app/models/eligibility_verification.rb
```diff
@@ -17,24 +30,38 @@ def self.expiration_warning
...
   private
 
-  def temporary_passenger
-    return if passenger.temporary?
+  def passenger_requires_validation
+    passenger&.active? && passenger&.temporary?
+  end
 
-    errors.add :base, 'must belong to a temporary passenger'
+  def wipe_contact_info
```
werebus: This method name is a little frightening, it mostly doesn't "wipe" the contact info.

(Not that I can think of a _better_ name)

app/models/eligibility_verification.rb
werebus: Although, on second thought, it really is optional.  Two opinions — that are a little at odds — I guess:

1.  The users are never presented with the opportunity to, I guess, type in an agency id, so it will always be one in the list. Then the only concern is what happens if an agency is deleted. Perhaps a `dependent: restrict` is warranted?
2.  Maybe it's a good idea to also add `validates :verifying_agency, etc`?

app/views/passengers/_form.html.haml
```diff
@@ -24,18 +27,29 @@
...
             - expiration_help = "Rides expire #{@passenger.rides_expire.try(:strftime, '%m/%d/%Y')}"
           = d.text_field :expiration_date,
             class: 'datepicker verification-expires',
-            help: expiration_help
+            help: expiration_help,
+            disabled: @passenger.permanent?
+          = d.form_group :contact_information,
+            label: { text: 'This verifying agency requires contact information to be entered' },
+            class: contact_information_class(@verification),
+            help: 'either phone or address is required' do
```
werebus: Both is fine, too?

db/migrate/20191122225846_make_verifying_agencies.rb
```diff
@@ -0,0 +1,13 @@
+class MakeVerifyingAgencies < ActiveRecord::Migration[5.2]
+  def up
+    %w[UHS Disability\ Services Accessible\ Workplace].each do |name|
```
werebus: When more than half of the "words" have spaces in them, I prefer a regular array literal. This way only saves you 3 characters. And, because the `\`es look like dividers, I parsed this as, "UHS Disability, Services Accessible, Workplace," on the first pass.

app/views/passengers/_contact_information.haml
```diff
@@ -0,0 +1,7 @@
+= verification.name
```
werebus: The file name of this partial might lead me to believe that it contained the contact information of the passenger.

app/assets/javascripts/passengers/passengers.js
```diff
@@ -19,4 +11,16 @@ $( document ).on("turbolinks:load", function() {
...
   });
+
+  $('#passenger_permanent').change(function(){
+    var expirationField = $('.verification-expires');
+    var permanent = $(this).is(':checked')
+    expirationField.prop('disabled', permanent);
+    if(permanent) { expirationField.val('') }
+  });
+
+  $('#passenger_eligibility_verification_attributes_verifying_agency_id').change(function(){
```
werebus: Whew, what a name.

app/models/eligibility_verification.rb
Anbranin: I did validate the verifying agency id under certain conditions. I think the only concern I have is someone deleting it, so I'll add the dependent: restrict.
If some yahoo gets hacky and adds a verifying agency ID that doesn't exist I won't lose sleep over it.

app/views/passengers/_form.html.haml
Anbranin: Yes but how do I say that without being way too wordy?
I do think a reasonable human would go with "well both is probably fine" rather than reading "but not both" from what's written... whatever that's worth.

werebus: Yeah, you're right.

app/assets/javascripts/passengers/passengers.js
Anbranin: I mean it was either that or *gasp* make my own ID when one already existed.. I don't know.

werebus: Yeah, I think this is fine; it's just a mouthful. The only other thing that came to mind was something like:

```
$("[id$='verifying_agency_id']")
```

But I don't think exchanging wordiness for more arcane symbols is a move in the positive direction.

werebus:
Two possible avenues on Codeclimate's complaint:

1.  Private controller methods, `#archive` and `#unarchive`.  Although, I think the "redirect and return" thing will be messy.
2.  A model method (`.toggle_archived`?) that does the assigning, saving, updating, etc. You could then just check whether the `User` is valid and act accordingly in the controller action.

Anbranin:
I think I did all the things

Anbranin:
> model method (.toggle_archived?) that does the assigning, saving, updating, etc. You could then just check whether the User is valid and act accordingly in the controller action.

I literally did exactly that before I even saw your suggestion GREAT MINDS THINK ALIKE EH?

PR #205, opened 2020-01-14
Fix migrations

Anbranin:
The migrations I wrote were in the wrong order:

ReplaceRegisteredWithDisabilityServicesWithVerifyingAgency runs. It saves EligibilityVerification objects. Those objects use a `before_save` block that references a column `name` that doesn't exist yet.

This should fix things.


werebus:
Unfortunately, this needs to be run before you remove the `registered_with_disability_services` column.

werebus:
Whoops. Nevermind.

Anbranin:
Yeah I did notice that and this _should_ be in the right order.

PR #207, opened 2020-01-22
Use after_create blocks to associate certain objects in FactoryBot

Anbranin:
For example, if passenger has an association :eligibility_verification
and eligibility verification has an association with passenger,
creating a passenger with an eligibility verification will create a
second passenger.
Tests pass and objects are properly related to each other with no
extras.

### DIFF
--- a/spec/factories/passengers.rb
+++ b/spec/factories/passengers.rb
@@ -7,7 +7,6 @@ FactoryBot.define do
     email { FFaker::Internet.email }
     phone { FFaker::PhoneNumber.short_phone_number }
     sequence(:spire) { |n| n.to_s.rjust(8, '0') + '@umass.edu' }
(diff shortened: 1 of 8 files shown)


spec/factories/passengers.rb
```diff
@@ -7,40 +7,55 @@
     email { FFaker::Internet.email }
     phone { FFaker::PhoneNumber.short_phone_number }
     sequence(:spire) { |n| n.to_s.rjust(8, '0') + '@umass.edu' }
+    # active_status { 'active' }
```
werebus: Commented-out?

Anbranin: Sometimes I comment out lines to see whether they make a difference or not. This time, I forgot to remove it when it didn't.

PR #213, opened 2020-03-25
Update Rails

werebus:


Anbranin:
Sometimes I think turbolinks is more trouble than it's worth.

werebus:
I'm sorry, I don't understand.

Anbranin:
No bearing on this PR I just can't say things out loud so I'm sprinkling my little nuggets of wisdom anywhere I can write text

PR #214, opened 2020-04-21
Cleanup passenger model

Anbranin:
Rubocop Passenger model and move things around a tiny bit.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -34,6 +34,12 @@ en:
   activerecord:
     errors:
       models:
+        passenger:
+          attributes:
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -34,6 +34,12 @@ en:
   activerecord:
     errors:
       models:
+        passenger:
+          attributes:
+            eligibility_verification:
+              blank: 'is required for temporary passengers with active registration'
+            spire:
+              invalid: 'must be 8 digits followed by @umass.edu'
```
avacmeyer, line 42: nice

PR #215, opened 2020-04-21
change toggling status to be more general

Anbranin:
now that there are three states a passenger can be in, make the method able
to change between any of those states


werebus:
Arguably not "toggling" (switching between two things)

Anbranin:
> Arguably not "toggling" (switching between two things)

I was actually wondering about that. I'll figure out something better.

app/views/passengers/archived.html.haml
```diff
@@ -24,7 +24,12 @@
           %td= passenger.expiration_display
           %td= link_to 'View', passenger
           %td= link_to 'Edit', edit_passenger_path(passenger)
-          %td= button_to 'Reactivate', toggle_archive_passenger_url(passenger), class: 'btn btn-warning'
-          %td= button_to 'Delete', passenger, method: :delete, data: { confirm: 'Are you sure?' }, class: 'btn btn-danger'
+          %td= button_to 'Reactivate',
+            toggle_status_passenger_url(passenger, status: 'active'),
```
werebus, line 28: `_path`

app/views/passengers/index.haml
```diff
@@ -50,7 +50,8 @@
           %td.text-center= passenger.rides_expire.try(:strftime, '%m/%d/%Y')
           %td= link_to 'View', passenger
           %td= link_to 'Edit', edit_passenger_path(passenger)
-          %td= button_to 'Archive', toggle_archive_passenger_url(passenger), class: 'btn btn-warning'
+          %td= button_to 'Archive', toggle_status_passenger_url(passenger, status: 'archived'),
```
werebus, line 53: `_path`

Anbranin:
Oh god. I settled on "change_status" but it isn't necessarily a change depending on what arguments you give it, is it?

app/views/passengers/archived.html.haml
Anbranin: Sure, but why? I know why I'd use url instead of path in some cases, but not any reasoning not to use path over url.

Anbranin: Plus `_url` is all over this site in the views. If there's a reason to switch, I can do all of them in a PR.

werebus: The reasons are minor, but it creates a dependency on Rails configuration that doesn't really need to be there. A link to "`/widgets`" is just as good as a link to "`http://mysite.example.com/widgets`", but one requires various values to be configured in the `config/` directory to work. There are also situations (admittedly misconfigurations) where the `_url` can conflict with e.g. your Apache configuration and result in needless redirects for every request.

A few people in the internet suggest "file size" as an argument, that seems ... weak.

I guess my bias comes from authoring plain-ol' HTML wobsites where being overly specific in your `href`s would eventually bite you (like when you decide to rename a directory, or add a sub-domain).

Anbranin: OK, I buy it.

PR #216, opened 2020-04-21
remove umass status from passenger

Anbranin:
Not sure why coverage went down, reviewing that now.


Anbranin:
Oh I get it, coverage went down because there are simply fewer lines in Passenger. So, not indicative of a problem.

Tanvish:
The changes look alright, I don't see an issue since they reflect the PR title. I just don't understand the logistics behind removing umass status. Is there an issue I need to look at?

Anbranin:
Thanks for the question, Tanvish; I forgot to mention that the ST Admin folks have found that they don't use this field and don't need it. I had also forgotten to make an issue about it although it was on my personal to-do list.

PR #217, opened 2020-04-22
add more granular restrictions.

Anbranin:
in preparation for ability to log in as a passenger.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -28,10 +28,7 @@ class ApplicationController < ActionController::Base
       elsif request.env.key? 'fcIdNumber'
         User.active.find_by spire: request.env['fcIdNumber']
       end
-    if @current_user.present?
-      session[:user_id] = @current_user.id
-    else redirect_to unauthenticated_session_path
-    end
+    session[:user_id] = @current_user.id if @current_user.present?
   end
   # rubocop:enable AbcSize
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,9 +2,9 @@
 
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+  before_action :check_primary_account
   before_action :set_current_user
   before_action :restrict_to_employee
-  before_action :check_primary_account
 
   private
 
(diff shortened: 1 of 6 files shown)


app/controllers/feedback_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class FeedbackController < ApplicationController
-  skip_before_action :access_control
+  skip_before_action :restrict_to_employee
```
werebus, line 4: `Feedback` currently rely on `user` being present in order to submit. Minor change required in `#signed_description`.

app/controllers/application_controller.rb
```diff
@@ -3,7 +3,7 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   before_action :set_current_user
-  before_action :access_control
+  before_action :restrict_to_employee
   before_action :check_primary_account
```
werebus, line 7: Minor optimization: order might be better as

```ruby
before_action :check_primary_account
before_action :set_current_user
before_action :restrict_to_employee
```

No need for a database lookup if we aren't going to let them in anyways.

app/controllers/application_controller.rb
```diff
@@ -3,7 +3,7 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   before_action :set_current_user
-  before_action :access_control
+  before_action :restrict_to_employee
```
werebus, line 6: Note that this action can't do anything yet because `#set_current_user` redirects to `SessionsController#unauthenticated` when `@current_user` isn't `present?`

app/controllers/feedback_controller.rb
Anbranin: Interesting, my dummy_feedback seemed to work but perhaps it didn't _really_ work.

werebus: Nope. I'm wrong. Skimmed too fast.

Anbranin: Actually, sorry, what won't work about it? It seems like line 55 `return description unless user.present?` would indicate that you wrote it such that it could work for people that weren't signed in.

app/controllers/application_controller.rb
Anbranin: It's changed in a later branch. But I may as well tweak set_current_user now.

app/controllers/feedback_controller.rb
Anbranin: I should reload before I comment :D

PR #220, opened 2020-05-19
Backend mailing code written, placeholder email template given

### LINKED ISSUE #156: Send email automatically when someone gets archived
The ST admins plan on notifying a passenger when they're being archived. I said "why don't I just do that automatically?" and they were like "oh... cool"

BenMelz:
closes #156

Added automatic emailing behavior when a passenger is archived.

As it stands the mailer is specific to passenger mail. I chose this over a more general class; in my mind it allows for more organization/scalability in the future with the downsides mainly being some unnecessary categorization.

The email is sent from the Passenger model, within the toggle_archived_status method. I couldn't really find an elegant way to implement sending the email while preserving the original return value. I took the route of making the changes uglier in favor of readability.

### DIFF
--- a/app/views/passenger_mailer/notify_archived.haml
+++ b/app/views/passenger_mailer/notify_archived.haml
@@ -1,3 +1,4 @@
-#{@passenger.name},
+Dear #{@passenger.name},
 
-Your account at Special Transit has been archived.
+Our records indicate that you only needed rides until #{Time.now.strftime("%m/%d/%Y")}, so we are archiving your account.
+If you do still need our service feel free to call us at (413) 545-2086.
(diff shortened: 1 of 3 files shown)


app/views/passenger_mailer/notify_archived.haml
```diff
@@ -0,0 +1,3 @@
+#{@passenger.name},
+
+Your account at Special Transit has been archived.
```
Anbranin, line 3: You should set your text editor so it puts a newline at the EOF. Read [this](https://thoughtbot.com/blog/no-newline-at-end-of-file) for more information.

app/views/passenger_mailer/notify_archived.haml
```diff
@@ -0,0 +1,3 @@
+#{@passenger.name},
```
Anbranin, line 1: I left specific wording for this on the issue, which you've seen.

PR #222, opened 2020-06-15
Creating new passenger now correctly records the registerer

### LINKED ISSUE #209: Record "registerer" when registering a passenger
Hello. I needed to see which dispatcher registered a passenger today because there was a mistake with the date. When I looked to see who it was, it said "Registering Dispatcher: Unknown". I noticed that it says this for all passengers. If we could see which dispatchers are registering passengers that would be super helpful!

[71](https://st-pax.admin.umass.edu/users/71)

BenMelz:
Closes #209 

As it was, the controller method was trying to assign the registerer by assigning `.registered_by` as the user's ActiveRecord object, when `.registered_by` is actually the foreign_key id column. 

As it was, displaying a passenger also trying to display the object string format (<ActiveRecord: ...>. Now it correctly displays the dispatcher's name.

### DIFF
--- a/app/views/passengers/show.haml
+++ b/app/views/passengers/show.haml
@@ -35,7 +35,7 @@
     = @passenger.rides_expire.strftime('%m/%d/%Y')
 %p
   %b Registering Dispatcher:
-  = @passenger&.registerer&.name
+  = @passenger.registerer&.name
 %p
   %b Registration Date:
   = @passenger.registration_date
--- a/app/views/passengers/show.haml
+++ b/app/views/passengers/show.haml
@@ -35,7 +35,7 @@
     = @passenger.rides_expire.strftime('%m/%d/%Y')
 %p
   %b Registering Dispatcher:
-  = @passenger.registerer ? @passenger.registerer.name : 'Unknown'
+  = @passenger&.registerer&.name
 %p
   %b Registration Date:
   = @passenger.registration_date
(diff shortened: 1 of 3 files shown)


app/views/passengers/show.haml
```diff
@@ -35,7 +35,7 @@
     = @passenger.rides_expire.strftime('%m/%d/%Y')
 %p
   %b Registering Dispatcher:
-  = @passenger.registerer || 'Unknown'
+  = @passenger.registerer ? @passenger.registerer.name : 'Unknown'
```
Anbranin, line 38: Here's a good place to use `try` or the [safe navigation operator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/)

app/views/passengers/show.haml
```diff
@@ -35,7 +35,7 @@
     = @passenger.rides_expire.strftime('%m/%d/%Y')
 %p
   %b Registering Dispatcher:
-  = @passenger.registerer ? @passenger.registerer.name : 'Unknown'
+  = @passenger&.registerer&.name
```
Anbranin, line 38: In this case we know that `@passenger` isn't going to be nil, so you don't need the safe operator on that one. While it's true that it won't _hurt_, it's always a good idea to trust the code. That is, if `@passenger` was indeed `nil` we'd want to know, and loudly.

BenMelz: Oh I see I think I got mixed up on what object each &. was operating on

PR #223, opened 2020-06-17
Fixed error handling for dispatch logs

### LINKED ISSUE #175: Submitting a blank dispatch log silently fails
I mean, maybe that's fine?  It seems unlikely that you'd leave the only field in a form blank.

But right now there _is_ error handling in the controller that doesn't do anything.

BenMelz:
Closes #175 

The server was 500ing due to some improper render calls on error, which I just combined with the success redirect since they serve the same purpose.

Also required the edit text box to have text on an html level. Now a user shouldn't be able to submit a bad request and if a bad request somehow happens, it will flash an error and not silently fail.

### DIFF
--- a/app/controllers/log_controller.rb
+++ b/app/controllers/log_controller.rb
@@ -7,11 +7,10 @@ class LogController < ApplicationController
     @entry = LogEntry.new entry_params.merge(user: @current_user)
     if @entry.save
       flash[:success] = 'Log entry was successfully created.'
-      redirect_to log_index_path
     else
       flash[:danger] = @entry.errors.full_messages
-      render :index
     end
+    redirect_to log_index_path
   end
 
   def destroy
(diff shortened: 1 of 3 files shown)


app/controllers/log_controller.rb
```diff
@@ -7,11 +7,10 @@ def create
     @entry = LogEntry.new entry_params.merge(user: @current_user)
     if @entry.save
       flash[:success] = 'Log entry was successfully created.'
-      redirect_to log_index_path
     else
       flash[:danger] = @entry.errors.full_messages
-      render :index
     end
+    redirect_to log_index_path
```
Anbranin, line 13: Yeah generally we render on failure because redirecting causes the user to lose all of the form data they may have filled out.
However, in this case, there's only one thing that can possibly be wrong and if it is, it means you haven't filled anything out. So carry on.

PR #228, opened 2020-07-15
Brochure Mailer

### LINKED ISSUE #200: Automatically send the brochure to passengers when their registration status changes to "active"
Saves dispatchers time. Alternatively, link to the brochure.

jpoleary:
Adds mailer that sends the brochure when a passenger's account is marked "activated". Pretty sure link is working correctly, but would appreciate someone double-checking it.

Closes #200


spec/mailers/passenger_mailer_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    it 'emails correct passenger' do
+      expect(output.to).to eql Array(@passenger.email)
+    end
+
+    it 'has correct subject' do
+      expect(output.subject).to eql 'Your UMass Transit Account'
+    end
+
```
Anbranin, line 29: I believe you can test the body of the output to make sure it contains the link you set up.

app/mailers/passenger_mailer.rb
```diff
@@ -3,4 +3,14 @@ def notify_archived(passenger)
...
   end
+
+  def send_brochure(passenger)
+    @passenger = passenger
+    mail to: @passenger.email, subject: 'Your UMass Transit Account'
+  end
+
+  def notify_pending(passenger)
+    @passenger = passenger
+    mail to: @passenger.email, subject: 'UMass Transit Account Created'
```
Anbranin, line 14: I just noticed the subject line--while the spec trans people may work out of the same building as Umass Transit, the department is actually "Special Transportation".
So the subject line should read "Special Transportation Account Created"

app/mailers/passenger_mailer.rb
```diff
@@ -3,4 +3,14 @@ def notify_archived(passenger)
     @passenger = passenger
     mail to: @passenger.email, subject: 'Special Transit Account Archived'
   end
+
+  def send_brochure(passenger)
+    @passenger = passenger
+    mail to: @passenger.email, subject: 'Your UMass Transit Account'
```
Anbranin, line 9: "Special Transportation Account Confirmed"

app/models/passenger.rb
```diff
@@ -29,6 +29,13 @@ class Passenger < ApplicationRecord
     if active_status_changed? && archived?
       PassengerMailer.notify_archived(self).deliver_now
     end
+    if active_status_changed? && active?
+      PassengerMailer.send_brochure(self).deliver_now
+    end
+  end
+
+  after_create do
+    PassengerMailer.notify_pending(self).deliver_now
```
Anbranin, line 38: When a passenger is created by a dispatcher (which is the only way it can happen right now), their status is not pending. We should only be sending the "your account is pending" email upon creation _if_ their account is pending.

Anbranin: I should modify what I said above to say that their account _can_ be active if created by a dispatcher--of course if the dispatcher selects "pending" as their status, that's not the case. But much of the time it is.

So if they get created and their account is active, they'll get two emails, which isn't ideal.

app/views/passenger_mailer/notify_pending.haml
```diff
@@ -0,0 +1,6 @@
+Dear #{@passenger.name},
+
+Your account with the Special Transportation Service has been created but is still pending. You can receive rides until (three business days from now, will insert actual date).
```
Anbranin, line 3: You should replace `(three business days from now, will insert actual date)` with the actual date three business days from whenever "now" is.

app/views/passenger_mailer/send_brochure.haml
```diff
@@ -0,0 +1,6 @@
+Dear #{@passenger.name},
+
+Hello and welcome to Special Transportation. A link to the brochure, where our terms and policies are outlined, is included below.
```
Anbranin, line 3: This verbiage should be:
> Dear (name),

> Your account with the Special Transportation Service has been activated. You can receive rides until (when rides expire). A link to the brochure, which includes our terms and policies, is included below.

> For more information,
(insert contact info)

app/models/passenger.rb
```diff
@@ -30,12 +30,16 @@ class Passenger < ApplicationRecord
...
     end
   end
 
   after_create do
-    PassengerMailer.notify_pending(self).deliver_now
+    if pending? do
+      PassengerMailer.notify_pending(self).deliver_now
+    elsif active? do
+      PassengerMailer.notify_active(self).deliver_now
+    end
```
Anbranin, line 42: Nice. Now write some tests for this.

app/views/passenger_mailer/notify_pending.haml
```diff
@@ -0,0 +1,9 @@
+Dear #{@passenger.name},
+
+Your account with the Special Transportation Service has been created but is still pending. You can receive rides until #{(Time.now+(60*60*24*3)).strftime("%m/%d/%Y")}.
+Once your account has been activated, we will notify you of the date until which you can receive rides. Before using our service, please read the brochure at our website at this link:
+
+=link_to passenger_url('brochure'), passenger_url('brochure')
+
+This is an automated email related to your UMass Amherst Transit Account.
+For more information, call us at (413)545-2086, email us at stspv@umass.edu, or visit our website at www.umass.edu/transportation/special-transportation
```
Anbranin, line 9: I like the signature! Nice job.

app/views/passenger_mailer/notify_active.haml
```diff
@@ -0,0 +1,9 @@
+Dear #{@passenger.name},
+
+Your account with the Special Transportation Service has been activated. You can receive rides #{('until ' + @passenger.rides_expire.strftime("%m/%d/%Y")) || 'permanently'}.
```
Anbranin, line 3: This line will never say "You can receive rides permanently". The `||` operator will only evaluate the second clause if the first one evaluates to either `nil` or `false`.

app/views/passenger_mailer/notify_pending.haml
```diff
@@ -0,0 +1,9 @@
+Dear #{@passenger.name},
+
+Your account with the Special Transportation Service has been created but is still pending. You can receive rides until #{(Time.now+(60*60*24*3)).strftime("%m/%d/%Y")}.
```
Anbranin, line 3: We want three business days from now. Also there are helpers (sprinkled all over the code for example) that will get you the day you want without you having to do any math.

app/views/passenger_mailer/notify_active.haml
```diff
@@ -0,0 +1,9 @@
+Dear #{@passenger.name},
+
+Your account with the Special Transportation Service has been activated. You can receive rides #{@passenger.rides_expire ? ('until ' + @passenger.rides_expire.strftime("%m/%d/%Y")) : 'permanently'}.
```
Anbranin, line 3: One small nitpick: "Permanent" passengers can't actually receive rides _forever_. They can receive them as long as they meet the requirements. So I think I would replace 'permanently' with 'until further notice'

Anbranin:
I think you missed something. Do a `git status`.

jpoleary:
> I think you missed something. Do a `git status`.

Not sure what I'm missing, but I also haven't had to use the schema yet (the rails tutorial didn't mention it at all) so I'm guessing it has something to do with that. Should I be putting through a migration to delete the table row as well? The only thing that git status gives me is coverage/.last_run.json

Anbranin:
I actually assumed you'd already made a migration and forgotten to add it because of your schema changes. Although now that I'm thinking about it, the schema would have had the new column if that was so, my bad.

Anyways yes, you need to remove the database column as well. We no longer need it.

Anbranin:
And you should definitely be adding the coverage file

db/migrate/20200729191001_remove_has_brochure_from_passengers.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveHasBrochureFromPassengers < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :passengers, :has_brochure, :boolean
```
Anbranin, line 3: Did you actually run this migration? I don't see a schema change.

jpoleary: I ran it and it said it successfully migrated, so I'm not sure why it wouldn't change.

werebus: https://github.com/umts/st-pax/pull/228/files#diff-1acd2e7e27a227829d5d14a91c863bb6L52

Anbranin: My mistake 👯‍♂️

PR #233, opened 2020-07-28
Updated capistrano/capistrano-bundler

werebus:
Also Bundler version, config updates. This should take care of the deprecation warnings from Bundler


werebus:
Travis failure addressed in #234.

PR #235, opened 2020-08-11
Navbar refactor

Anbranin:
There are a couple other pages I want to add to the navbar, so these dropdowns will prevent it from overflowing on most screens, which doesn't look all that great.


Anbranin:
Should I made all the link text gray? Thoughts? It's all red right now which.. doesn't look amazing, but gray also seems odd.

Anbranin:
I actually am unable to get a consistent styling across all the navbar buttons now unless I DIY some CSS. I can't use bootstrap classes. If I apply the `btn.btn-outline-secondary` class to the link itself, the dropdown arrow is outside of the outline. If I apply the btn.btn-outline to the nav-link, it looks like it does with the link being red and not becoming a nice white on hover.
Am I wrong about there not being a nice bootstrap solution to this problem, like am I not thinking big enough (should I do a complete redesign instead of leaving some buttons as-is?) or should I write my own CSS to make those links look consistent with the others? I think it's a good idea not to leave them red?

werebus:
FWIW, Bootstrap's [navbar elements](https://getbootstrap.com/docs/4.5/components/navbar/) don't usually have "buttons" in them (or, at least not everything is a button — an actual form action is OK). That's just a thing we've done on a few of our sites.

I actually don't really like the look, myself.

Anbranin:
What don't you like the look of? Bootstrap's navbar? I think it's fine. I hate navbars in general so they're all the same to me.

werebus:
I don't like buttons in the bar.  Certainly not _all_ buttons. Why are they buttons? What does that offer over a link?

Anbranin:
Figured out a nice link solution. Not done yet - have to add back in logic about who can see what, and decide whether or not to remove the helper method or just change it.

Anbranin:
⚡

app/views/layouts/_navbar.haml
```diff
@@ -1,12 +1,28 @@
...
+    - if @current_user
+      %li.nav-item.dropdown
+        = link_to 'Passengers', '#', class: 'nav-link dropdown-toggle', id: 'passenger-dropdown',
+          data: {toggle: 'dropdown'}, aria: {haspopup: 'true', expanded: 'false'}
+        .dropdown-menu{aria: { labelledby: 'passenger-dropdown' }}
+          = link_to 'Current', passengers_path, class: 'dropdown-item'
+          = link_to 'Archived', archived_passengers_path, class: 'dropdown-item'
+          = link_to pending_passengers_path, class: 'dropdown-item' do
+            Pending
+            %span.badge.badge-warning= Passenger.pending.count
```
werebus, line 12: I think this badge will be a bit more effective if it isn't displayed when there are 0.

app/views/layouts/_navbar.haml
```diff
@@ -1,12 +1,28 @@
...
-    = navbar_link 'Manage Users', users_path
-    = navbar_link 'Manage Mobility Devices', mobility_devices_path
-  = navbar_link 'Site Feedback', feedback_index_path
-  = navbar_link 'Brochure', brochure_passengers_path
-  = link_to 'Logout', destroy_session_path, class: 'ml-auto mr-2 nav-link btn btn-secondary'
+%nav.navbar.navbar-expand-lg.navbar-light.bg-light{aria: {label: 'Main'}}
+  %ul.navbar-nav.mr-auto
+    - if @current_user
+      %li.nav-item.dropdown
+        = link_to 'Passengers', '#', class: 'nav-link dropdown-toggle', id: 'passenger-dropdown',
```
werebus, line 5: Any though of trying to add some sort of signal here that there are pending pax?

werebus (changes requested):
Style looks good to me. Just one minor thing:

app/views/layouts/_navbar.haml
Anbranin: *sigh* yeah probably the same badge.

Anbranin: Nope, that looks dumb and doesn't fit. highlight yellow?

Anbranin: It looks so gross. What do

Anbranin: I have settled on a font-awesome badge. Taking suggestions.

app/views/layouts/_navbar.haml
```diff
@@ -1,12 +1,34 @@
...
+  %ul.navbar-nav.mr-auto
+    - if @current_user
+      %li.nav-item.dropdown
+        = link_to '#', class: 'nav-link dropdown-toggle',
+          id: 'passenger-dropdown',
+          data: {toggle: 'dropdown'},
+          aria: {haspopup: 'true', expanded: 'false'} do
+          Passengers
+          - unless Passenger.pending.count.zero?
+            %i.fas.fa-user-cog{aria_label: 'Pending passengers present'}
```
werebus, line 11: The "cog" in the icon says to me "settings" or "preferences". For the concept of "pending", I think I prefer the clock (implying that they're waiting) or perhaps the "+"?

Also, I would tie it to the internal badge by color (add `.text-warning`)

werebus (changes requested):

app/views/layouts/_navbar.haml
Anbranin: I like the clock better. The plus doesn't indicate, to me, something that you could make disappear by doing some actions.

werebus (approved):

PR #238, opened 2020-08-17
Rescue smtp error

### LINKED ISSUE #236: Archive Passenger Error
I am trying to archive a permanent passenger and I am getting an error message. It says:
We're sorry, but something went wrong.
If you are the application owner check the logs for more information.

[71](https://st-pax.admin.umass.edu/users/71)

BenMelz:
Closes #236 

Pretty self explanatory. Warn users if the update email doesn't send.
Turns out raising an exception in a `before_save` block doesn't prevent the save from executing, so what I wrote this morning will work fine.


Anbranin (dismissed):
This looks good so far but please write tests for it.

BenMelz:
I knew someone would say that right after I put the PR up

Tests foil me once again!

BenMelz:
Good thing you made me write those, turns out `before_save` and `after_save` are indeed transactional and whatever I did in the console to come to the conclusion that they were not was not right. `after_commit` is what we're looking for.

app/controllers/passengers_controller.rb
```diff
@@ -14,12 +14,18 @@ def archived
...
+      if @passenger.set_status(params[:status])
+        flash[:success] = 'Passenger successfully updated'
+        redirect_to passengers_url
+      else
+        flash[:danger] = @passenger.errors.full_messages
+        redirect_to edit_passenger_path(@passenger)
+      end
+    rescue Net::SMTPFatalError
+      flash[:warning] = 'Passenger successfully updated, ' \
+                        'but the passenger was unable to be notified of their status change via email.'
```
werebus, line 27: A bit too much [passive voice][srbm] for me. Perhaps, "but we could not notify them of their status change" 

[srbm]: https://youtu.be/v-8_LWq236Y

spec/controllers/passengers_controller_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  let!(:admin) { create :user, :admin }
+  before :each do
+    session[:user_id] = admin.id
+  end
+
+  describe 'POST set_status' do
+    context 'with broken mailer' do
+      context 'changing active to archived' do
+        it 'saves the passenger and displays a warning message' do
+          allow(PassengerMailer).to receive(:notify_archived).and_raise Net::SMTPFatalError
```
werebus, line 16: I would move this to a `before` block inside `'with a broken mailer'`. It's setup that defines that context.

spec/controllers/passengers_controller_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+          @passenger = create :passenger, :permanent, active_status: 'active'
```
werebus, line 18: `@passenger` doesn't need to be an instance variable. if it's only used in this block.

That said, maybe it _should_ be in a `let` block?

spec/controllers/passengers_controller_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+          post :set_status, params: { id: @passenger.id, status: 'archived' }
+          @passenger.reload
+
+          expect(@passenger.active_status).to eq 'archived'
+          expect(response).to redirect_to(passengers_path)
+          expect(flash[:warning]).to match(/the passenger was unable to be notified of their status change via email/)
```
werebus, line 24: I'm always a little wary of super-log regexes for UI strings. Pare it down a bit to the essence of the warning?

spec/controllers/passengers_controller_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  end
+
+  describe 'POST create' do
+    context 'with broken mailer' do
+      context 'creating a pending passenger' do
+        it 'saves the passenger and displays a warning message' do
+          allow(PassengerMailer).to receive(:notify_pending).and_raise Net::SMTPFatalError
+
+          new_passenger_params = {
+            passenger: {
```
werebus, line 50: You can do `attributes_for :passenger, :permanent` to make Factory Bot build this hash.

spec/controllers/passengers_controller_spec.rb
```diff
@@ -9,91 +9,84 @@
...
-          @passenger.reload
+          passenger.update_columns active_status: 'active'
+          post :set_status, params: { id: passenger.id, status: 'archived' }
+          passenger.reload
 
-          expect(@passenger.active_status).to eq 'archived'
+          expect(passenger.active_status).to eq 'archived'
           expect(response).to redirect_to(passengers_path)
-          expect(flash[:warning]).to match(/the passenger was unable to be notified of their status change via email/)
+          expect(flash[:warning]).to match(Regexp.union(/could not notify/, /status change/))
```
werebus, line 28: `Regexp.union` (more or less) creates an alternation. That is, this is the same as `/could not notify|status change/`, which will match if either phase is present in the string. I suspect you wanted _both_. If so, the easiest way is probably two expectations.

spec/controllers/passengers_controller_spec.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe PassengersController do
+
+  let!(:admin) { create :user, :admin }
```
werebus, line 7: Don't need `!`

spec/controllers/passengers_controller_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  before :each do
+    session[:user_id] = admin.id
+  end
+
+  let!(:passenger) { create :passenger, :permanent }
```
werebus, line 12: Don't need `!`

werebus (changes requested):

werebus (approved):

PR #240, opened 2020-09-14
Update ruby to 2.6.6

jforgue:
st-pax shares a server with [strap](https://github.com/umts/strap) so that will need to be updated as well before this can be merged.


werebus:
Travis is broken.

Quick fix is to constrain `simplecov` to `<= 0.13` (https://github.com/codeclimate/ruby-test-reporter/pull/181).  Better fix is to set up CodeClimate's "new unified test reporter client".

werebus:
See #242

Gemfile
```diff
@@ -3,8 +3,10 @@
 source 'https://rubygems.org'
 ruby IO.read(File.expand_path('.ruby-version', __dir__)).strip
 
+# autoprefixer-rails 10 or greater requires node.js 12 or greater
+gem 'autoprefixer-rails', '~> 9'
 gem 'bootstrap', '~> 4.3.1'
-gem 'bootstrap_form'
+gem 'bootstrap_form', '~> 4.3.0'
```
jforgue, line 9: `bootstrap_form >= 4.5` depends on `autoprefixer-rails >= 10`. `autoprefixer-rails >= 10` requires `node.js >= 12`. Currently the server is running `node.js 11.15.0`.

werebus: So, I don't, um, see this to be the case. There's no mention of that requirement in `bootstrap_form`'s README or gemspec.

Also — minor correction — our server is running node.js 10.15.2

Gemfile
```diff
@@ -3,8 +3,10 @@
 source 'https://rubygems.org'
 ruby IO.read(File.expand_path('.ruby-version', __dir__)).strip
 
+# autoprefixer-rails 10 or greater requires node.js 12 or greater
+gem 'autoprefixer-rails', '~> 9'
```
werebus, line 7: This works because Bundler tries to be smarter than you, but:

* `'~> 1.1.1'` means 1.1.1 or 1.1.2 or 1.1.78, but not 1.2.x.
* `'~> 1.1'` means 1.1.x or 1.2.x or 1.100.x but not 2.x.y.
* By extension, you would think `'~> 9' would mean '9.x or 10.x or 67.x, but no version larger than infinity.

Luckily Bundler thinks that's dumb. `'~> 9'` and `'~> 9.0'` mean the same thing; but I prefer the one that follows the same pattern as the others.

werebus (changes requested):
I'm pretty sure you don't need the `bootstrap_form` constraint.

werebus:
Closing in favor of #277

PR #242, opened 2020-09-16
Attempt to use the new CodeClimate test reporter

werebus:


Anbranin (approved):
I read the docs and see nothing amiss!

werebus:
Yeah. Turns out it doesn't.... [work per se](https://github.com/codeclimate/test-reporter/issues/413). But, it will soonish and also our current setup has never successfully reported coverage to CodeClimate.

PR #243, opened 2020-09-16
Another date fix

werebus:
As soon as the passenger is saved, their registration _date_ will be assigned to be `Time.zone.today`, so it doesn't make sense for it to be a date and time before the passenger is saved.

Also, it should be 3 business days after the beginning of today, not the beginning of the day 3 business days from now.


Anbranin (approved):
Took me a hot second to figure out what you meant but this makes sense.

Issue #231, opened 2020-07-21
Add ability to subscribe to text messages to passengers


Anbranin:
Some passengers would like to receive communication via text versus over the phone, as is done currently. It sure would be nice from an accessibility standpoint to be able to text passengers in an official manner (and make sure they're opted-in). The solution decided upon is that the dispatchers or admins can use email to send texts to passengers when that feature is available.

This means that we need to add two columns to Passenger: A boolean like "subscribed_to_sms" with a default of false, and a carrier_id related to a new model "Carrier". (or CellCarrier, whatever).

A Carrier should have a gateway address and a name. An example is Verizon. The name is "Verizon" and the gateway address is "vztext.com". To text a Verizon number, you'd email `[10-digit cell number]@vztext.com`.

A page to manage carriers is required here, basically a scaffold, maybe some help text indicating how things work would be nice. An additional nicety would be to construct an email address for each passenger based on their cell number and gateway and put that on their show page. It's certainly possible.

Make sure that a couple example cell carriers are added to the seeds file.

PR #247, opened 2020-10-13
Add subscribe to sms for passengers

jforgue:
I should probably add more tests for the js and passenger validations, but I would like to know how everyone feels about the views.


app/views/carriers/index.haml
```diff
@@ -0,0 +1,25 @@
...
+              %td= carrier.gateway_address
+              %td= link_to 'Edit', edit_carrier_path(carrier)
+              %td.text-right= button_to 'Destroy', carrier, method: :delete,
+                class: 'btn btn-danger', data: { confirm: 'Are you sure?' }
+        - else
+          %tr
+            %td{colspan: 7} No carriers to show.
+    .row
+      .col.text-right
+        = link_to 'New Carrier', new_carrier_url
```
Anbranin, line 25: I think we slightly prefer `path` but I can't remember why. @werebus may remember

db/migrate/20200902174022_add_subscribed_to_sms_to_passengers.rb
```diff
@@ -0,0 +1,11 @@
+class AddSubscribedToSmsToPassengers < ActiveRecord::Migration[5.2]
+   def up
+    add_column :passengers, :subscribed_to_sms, :boolean, default: false
+    add_column :passengers, :carrier_id, :text
```
Anbranin, line 4: an ID column should be an integer.

spec/factories/passengers.rb
```diff
@@ -56,5 +56,10 @@
     trait :permanent do
       permanent { true }
     end
+
+    trait :subscribed_to_sms do
+      subscribed_to_sms { true }
+      carrier { Carrier.all.sample }
```
Anbranin, line 62: Carriers won't already exist here. `association :carrier` is correct.

spec/system/carriers_management_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      it 'successfully deletes the carrier' do
+        page.accept_confirm 'Are you sure?' do
+          click_button 'Destroy'
+        end
+        expect(page).to have_text 'Carrier successfully destroyed'
+        within 'tbody' do
+          expect(page).to have_text 'No carriers to show.'
+        end
+      end
+      it 'fails to delete the carrier' do
```
Anbranin, line 91: I don't understand... what causes a carrier to not be able to be destroyed? A context is needed here.

spec/system/carriers_management_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+        submit
+        expect(page).to have_text 'Carrier successfully updated'
+        within 'tbody' do
+          expect(page).to have_text 'Verizon'
+          expect(page).not_to have_text 'tmobile'
+          expect(page).to have_text '@vzw.com'
+          expect(page).not_to have_text '@tmo.com'
+        end
+      end
+      it 'fails to update the carrier' do
```
Anbranin, line 70: You should use `context`s here or even just add a "when name is left blank" to your test description. It helps readability to know under what conditions a carrier fails to update without having to read the code.

Anbranin (changes requested):

app/views/carriers/index.haml
werebus: Yeah, the rationale is pretty weak, but it amounts to: "you don't need protocol, host, and port here, so why add that extra information to the page?" In days of yore, making links on your wobsite absolute really mucked things up when you went to move your static files to a hew host, a new subdirectory, etc. That  _doesn't_ apply here, but it's the background for my opinion.

spec/system/carriers_management_spec.rb
jforgue: There is no reason a carrier should fail to be destroyed, but that doesn't mean it won't happen (due to a server issue or something). We wrote the code to handle a destroy failing so it needed to be tested.

Anbranin: Oh no, my complaint wasn't about that. I just actually didn't see any difference between the two tests, but I missed the `allow_any_instance` you had below. Carry on.

app/views/carriers/_form.haml
```diff
@@ -0,0 +1,6 @@
+.row
+  .col-md-6.offset-md-3
+    = bootstrap_form_with model: @carrier, local: true do |f|
+      = f.text_field :name
+      = f.text_field :gateway_address
```
Anbranin, line 5: If I were filling this out, I'd wonder if I should put the `@` in this text field or if it'd be filled in. Later on, your code assumes that they do begin with the `@`. Could you validate that the gateway address begins with `@`?

Anbranin (changes requested):

Anbranin:
I changed the navbar quite significantly, but I can tell you that your carrier link should go under the "Manage" section and be viewable by dispatchers.

jforgue:
Closes #247.

jforgue:
Closes #231.

spec/system/passenger_management_spec.rb
```diff
@@ -10,48 +10,67 @@
...
+          create :carrier
+          date = 2.days.since.strftime '%Y-%m-%d'
+          visit passengers_path
+          click_on 'Add New Passenger'
+          fill_in 'Passenger Name', with: 'Foo Bar'
+          fill_in 'Email', with: 'foobar@invalid.com'
+          fill_in 'Address', with: '123 turkey lane'
+          fill_in 'Phone', with: '123'
+          # Carrier selectbox has no blanks
+          check 'Subscribed to sms'
```
Anbranin, line 32: Why do we need to check subscribed here? What do you mean by "carrier selectbox has no blanks"?
You should leave a test that creates a passenger with the absolute minimum necessary to create a passenger. That is, create one without being subscribed to text messages. 
Your first context should be "creating a new passenger that is subscribed to SMS", leaving the other success test alone.

app/views/layouts/_navbar.haml
```diff
@@ -27,6 +27,10 @@
         .dropdown-menu{aria: { labelledby: 'management-dropdown' }}
           = link_to 'Users', users_path, class: 'dropdown-item'
           = link_to 'Mobility Devices', mobility_devices_path, class: 'dropdown-item'
+          = link_to 'Carriers', carriers_path, class: 'dropdown-item'
+    - elsif @current_user&.dispatcher?
+      %li.nav-item
```
Anbranin, line 32: Let's show the "Manage" dropdown for everyone, and just show different links within to various people.

Anbranin (changes requested):
Great job, just a couple more thoughts

app/views/layouts/_navbar.haml
```diff
@@ -20,13 +20,15 @@
         = link_to 'Add New Passenger', new_passenger_path, class: 'nav-link'
       %li.nav-item
         = link_to 'Dispatch Log', log_index_path, class: 'nav-link'
-    - if @current_user&.admin?
+    - if @current_user
```
jforgue, line 23: Because passengers aren't users, `@current_user` is `nil` when a passenger is logged in. This seem like a strange choice to me.

jforgue:
It seems that occasionally some user_management_specs will fail when running the whole suit because `@current_user` isn't being set correctly. I'm not sure if this was already an issue and I'm not sure why it's happening. Even when I get a failure then run the same seed again it passes. Maybe it's just my environment?

app/views/layouts/_navbar.haml
Anbranin: If there's something actually wrong with how this is done, feel free to make an issue about it. Otherwise, the time to comment on it would have been the PR where it was added.

Anbranin:
I'll double check the tests on my machine

spec/factories/passengers.rb
```diff
@@ -56,5 +56,10 @@
     trait :permanent do
       permanent { true }
     end
+
+    trait :subscribed_to_sms do
+      subscribed_to_sms { true }
+      association :carrier
+    end
```
Anbranin, line 63: Now you aren't using this trait any more. I don't think it's useful to have as a trait tbh

app/views/passengers/show.haml
```diff
@@ -11,6 +11,14 @@
...
   %b Phone:
   = @passenger.phone
+%p
+  %b Subscribed to SMS:
+  = @passenger.subscribed_to_sms ? "Yes" : "No"
+
+- if @passenger.subscribed_to_sms?
+  %p
+    %b SMS address:
+    = @passenger.phone.delete('-') + @passenger.carrier.gateway_address
```
Anbranin, line 21: What if the phone number is saved as `(413).666.6666`, will this still work?
P.S. Please don't validate phone number formats, that's not what I'm getting at. Just change this line.

Anbranin (changes requested):

spec/factories/passengers.rb
jforgue: Oh, I forgot I made that.

app/views/passengers/show.haml
jforgue: I assumed phone numbers were already validated for some reason, I'll fix it.

app/views/layouts/_navbar.haml
jforgue: I wouldn't say anything is wrong, it just confused me. I guess I just wanted confirmation that I had the right idea.

app/views/passengers/show.haml
```diff
@@ -18,7 +18,7 @@
 - if @passenger.subscribed_to_sms?
   %p
     %b SMS address:
-    = @passenger.phone.delete('-') + @passenger.carrier.gateway_address
+    = @passenger.phone.tr('^0-9', '') + @passenger.carrier.gateway_address
```
Anbranin, line 21: til the `tr` method. huh.

Anbranin: out of curiosity, why not `gsub`?

Anbranin: OK never mind I looked it up, `tr` is faster but I didn't think it took regular expressions as an argument. Seems to work though.

jforgue: What was I thinking.

app/views/layouts/_navbar.haml
Anbranin: It _is_ different from our other apps, good on you for figuring out what was going on.

Anbranin (approved):

PR #248, opened 2020-11-02
Self registration

Anbranin:
MVP for self-registration. I need some feedback on these design decisions I've made. Hurt me.


Anbranin:
I have opened this PR before finishing the tests I'm writing, but don't worry, they're being written.

Anbranin:
I WROTE TESTS AND THEY'RE TREMENDOUS

app/assets/javascripts/passengers/passengers.js
```diff
@@ -36,6 +36,7 @@ $( document ).on("turbolinks:load", function() {
 
   $('#passenger_eligibility_verification_attributes_verifying_agency_id').change(function(){
     var needsContactInfo = $(this).children("option:selected").data('needs-contact-info');
+    needsContactInfo ||= false; // blank option will needlessly cause toggling
```
Tanvish, line 39: This is interesting and pretty cool.

app/controllers/application_controller.rb
```diff
@@ -31,6 +32,28 @@ def set_current_user
...
+        Passenger.find_by(id: session[:passenger_id])
+      elsif request.env['fcIdNumber']
+        find_or_initialize_passenger
+      end
+    session[:passenger_id] = @registrant&.id
+  end
+
+  def find_or_initialize_passenger
+    passenger = Passenger.find_or_initialize_by(
+      spire: "#{request.env['fcIdNumber']}@umass.edu",
```
Tanvish, line 49: Is spire really the `#######@umass.edu?` I thought it was simply the 8 digit number. Is this is a logistics thing?

Tanvish:
The form looks nice. The design is simplistic and minimal which is pretty good. The back and edit links look a bit awkward. But I wonder if it's that worth giving it more thought than you already have.

Screen Shot 2020-12-28 at 7 10 25 PM

The design is pretty good. There were some points for the form though.
- The phone number needs some sort of filter to make sure it is not invalid.
- The mobility device didn't make sense to me. Are you talking about the device linked with the phone number? Also, I don't see any options to choose from in that section.
- For the `How long will you be with us?` section, are we actually asking `How long will you need our services starting from today?`. I'm just curious as to how that actually works in real life.

For the code part, I'm just commenting on things that I don't understand but it's mostly looking good. I'll just a couple of questions here and there to make sure I fully understand everything.

jforgue (changes requested):
Some non-tremendous tests need to be fixed other than that I didn't see any issues but will rereview when all tests are passing

app/controllers/application_controller.rb
Anbranin: In this database it's saved with the `@umass.edu`. I don't remember why it was done that way.

Anbranin:
@jforgue I fixed the tests, just needed merge conflict resolution. LMK what you think now.

Anbranin:
@Tanvish  to address your points:

> The phone number needs some sort of filter to make sure it is not invalid.

I don't disagree, however this PR has almost nothing to do with phone numbers. It's supposed to be an MVP to get people started with self-registration and that's all. Furthermore, validating phone numbers isn't as easy as you would think. We have passengers with international numbers, possibly extensions, and there are a few valid ways to write a number. For this reason we haven't validated it yet and may not plan to.

> The mobility device didn't make sense to me. Are you talking about the device linked with the phone number? Also, I don't see any options to choose from in that section.

A "mobility device" refers to any device which allows a passenger to move better, such as a wheelchair, a walker, or a cane. The reason we ask is because some mobility devices indicate that the passenger needs more time slots per ride--that is, bringing them from place to place will take longer because we need to, for example, deploy the wheelchair lift for them. The reason you didn't see any options is because you hadn't run `rake db:seed`. Had you done so, the answer to this question might have become clear.

> For the How long will you be with us? section, are we actually asking How long will you need our services starting from today?. I'm just curious as to how that actually works in real life.

Yes, that's what we're asking. I got the verbiage from the SpecTrans supervisors--they told me what they wanted the fields to say. So saying "how long will you be with us?" is directly requested by the customer and while I _can_ change it, I would have to run it by them, and I think that in this case they might have the best idea of what to write.

The plan right now is to get this out and have them try it with a few passengers before announcing self-registration to the general public. A "soft opening" if you will, kind of like having it in beta. So once I get some feedback there will doubtless be more changes and possibly some bugs to fix.

spec/system/passenger_registration_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+        visit edit_passenger_path(@passenger)
+        expect(page).not_to have_field 'Address'
+        expect(page.current_url).to include passenger_path(@passenger)
+        expect(page.current_url).not_to include edit_passenger_path(@passenger)
+        expect(page).to have_text "To edit your profile, please call"
+        visit register_passengers_path
+        expect(page).not_to have_field 'Address'
+        expect(page.current_url).to include passenger_path(@passenger)
+        expect(page.current_url).not_to include edit_passenger_path(@passenger)
+        expect(page).to have_text "To edit your profile, please call"
```
jforgue, line 58: Single quotes
```suggestion
        expect(page).to have_text 'To edit your profile, please call'
```

spec/system/passenger_registration_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+    context 'after becoming active' do
+      it 'does not allow editing' do
+        @passenger = create :temporary_passenger, :with_note
+        login_as(@passenger)
+        visit edit_passenger_path(@passenger)
+        expect(page).not_to have_field 'Address'
+        expect(page.current_url).to include passenger_path(@passenger)
+        expect(page.current_url).not_to include edit_passenger_path(@passenger)
+        expect(page).to have_text "To edit your profile, please call"
```
jforgue, line 53: Single quotes
```suggestion
        expect(page).to have_text 'To edit your profile, please call'
```

spec/system/passenger_registration_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+        expect{ submit }.to change { Passenger.count }.by 1
+        expect(page).to have_text 'Passenger registration successful'
+        expect(Passenger.last).to be_pending
+      end
+    end
+    context 'unsuccessfully' do
+      it 'renders errors in the flash' do
+        visit register_passengers_path
+        fill_in 'Address', with: '123 turkey lane'
+        expect{ submit }.not_to change { Passenger.count }
```
jforgue, line 27: Missing a space before opening bracket
```suggestion
        expect { submit }.not_to change { Passenger.count }
```

jforgue (approved):
I don't see any big issues, just a few linting things I would recommend

Anbranin:
Why Travis why you do me dirty like that

app/assets/javascripts/passengers/passengers.js
jforgue, line 39: Chromium version 85 or newer is required to support the `||=` operator. I recommend you set this without this new fangled operator for comparability's sake. 
```suggestion
    needsContactInfo = needsContactInfo || false; // blank option will needlessly cause toggling
```

Anbranin: `||=` seems like such a basic operation that I'd have _never_ guessed it would be unsupported in even older browsers. Thanks for debugging this for me! <3

PR #254, opened 2021-01-11
Finish rails 5.2 upgrade

werebus:
The defaults have been running in prod for about 2 weeks now.


config/puma.rb
```diff
@@ -9,15 +9,17 @@
...
+# Specifies the `port` that Puma will listen on to receive requests; default is 3000.
 #
 port        ENV.fetch('PORT') { 3000 }
 
 # Specifies the `environment` that Puma will run in.
 #
 environment ENV.fetch('RAILS_ENV') { 'development' }
 
+# Specifies the `pidfile` that Puma will use.
+pidfile ENV.fetch('PIDFILE') { 'tmp/pids/server.pid' }
```
Anbranin, line 21: Do we have to add `!/tmp/pids/.keep` to the gitignore now or??

werebus: Nope, that's the default either way. If the directory doesn't exist, Puma creates it.

This is just the config that comes with a new Rails 5.2 app. It allows for configuration via environment variables which is particularly useful for production environments like Heroku. We don't use puma in production, though.

Anbranin: I 100% forgot we don't use it in production. I read that if the directory doesn't exist, Puma just doesn't start, but that article was clearly wrong.

Anbranin (approved):

config/puma.rb
werebus: To be fair, what you read is correct.  **Puma** won't start if the directory doesn't exist:

```
$bundle exec puma
Puma starting in single mode...
* Version 3.12.6 (ruby 2.6.1-p33), codename: Llamas in Pajamas
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
bundler: failed to load command: puma (/home/matt/.rbenv/versions/2.6.1/bin/puma)
Errno::ENOENT: No such file or directory @ rb_sysopen - tmp/pids/server.pid
```

The `rails server` command does some [additional smoothing-over](https://github.com/rails/rails/blob/2a3f8a29e5d7d17a0448fa47fd756c8d11b175b1/railties/lib/rails/commands/server/server_command.rb#L82-L86) for you.

PR #274, opened 2021-04-22
Combine pax views

### LINKED ISSUE #229: Use the same view for archived and active passengers
There's no reason "archived" can't be a filter just like temporary/permanent passengers. It's a good deal of work to restructure the view but I think it's worth it because the navbar is getting cluttered.. and it's about to get more cluttered once I add "pending passengers" to that list.

werebus:
Closes #229 

Use the same view for current, pending and archived passengers. But, keep the existing named routes.


Anbranin (approved):
great!

werebus:
Travis failed because of a (very small) coverage drop - likely because I removed some covered controller methods.

PR #280, opened 2021-05-17
Cache gems with Travis

werebus:
Also, ease up on the coverage drop requirement a bit, and fix CodeClimate to work with new SimpleCov


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "line": 82.81
+    "line": 82.78
```
werebus, line 3: This is why `refuse_coverage_drop` is bad in development. This drop happened who knows how long ago, but it wasn't written to the file because we told SompleCov not to.

Anbranin: SOMPLECOV

Anbranin (approved):

PR #286, opened 2021-06-02
Replaced active_status with registration_status, in…

### LINKED ISSUE #230: Change "active_status" to "registration_status"
"registration status" is a much better term.

jpoleary:
Changes the active_status column in passengers to registration_status for clarity purposes. Closes #230

### DIFF
--- a/db/migrate/20210602150743_change_active_status_to_registration_status.rb
+++ b/db/migrate/20210602150743_change_active_status_to_registration_status.rb
@@ -1,0 +1,8 @@
+class ChangeActiveStatusToRegistrationStatus < ActiveRecord::Migration[6.1]
+  def up
+    rename_column :passengers, :active_status, :registration_status
+  end
+  def down
+    rename_column :passengers, :registration_status, :active_status
+  end
+end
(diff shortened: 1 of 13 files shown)


db/migrate/20210602150743_change_active_status_to_registration_status.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeActiveStatusToRegistrationStatus < ActiveRecord::Migration[6.1]
+  def up
+    rename_column :passengers, :active_status, :registration_status
+  end
+  def down
+    rename_column :passengers, :registration_status, :active_status
+  end
```
Anbranin, line 7: I feel like you should embrace `change` to keep up with the times, but it doesn't really matter. What _does_ matter though is that I see that you didn't actually run this migration, or at least didn't commit your schema changes when you did!

Anbranin (changes requested):
I don't see any schema changes.

db/migrate/20210602150743_change_active_status_to_registration_status.rb
jpoleary: ahh whoops!

Anbranin (approved):

PR #288, opened 2021-06-03
Removed permanent checkbox

### LINKED ISSUE #285: Remove "permanent?" checkbox from passenger registration view
They can never edit the checkbox, it's disabled right now. Just change the logic so it's not visible.

jpoleary:
Removed the 'permanent' checkbox for passenger registration. Closes #285

### DIFF
--- a/app/views/passengers/_form.html.haml
+++ b/app/views/passengers/_form.html.haml
@@ -23,6 +23,10 @@
         options_from_collection_for_select(MobilityDevice.order(:name),
           :id, :name, passenger.mobility_device&.id), include_blank: 'None'
 
+      - if @current_user&.admin?
+        = form.form_group do
+          = form.check_box :permanent
+
       = form.form_group do
         = form.check_box :subscribed_to_sms
 
--- a/app/views/passengers/_form.html.haml
+++ b/app/views/passengers/_form.html.haml
@@ -15,16 +15,13 @@
         = form.form_group :registration_status,
           label: { text: 'Registration Status' },
           help: "You can select 'pending' if the passenger is temporary but has no verifying information yet" do
-          = form.radio_button :active_status, 'pending', class: 'form-check-input'
-          = form.radio_button :active_status, 'active', class: 'form-check-input'
-          = form.radio_button :active_status, 'archived', class: 'form-check-input'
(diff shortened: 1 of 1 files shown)


jpoleary:
Should I also remove the 'permanent' section for passenger display, or would that still be useful?

Anbranin:
Displaying it is fine because it's pertinent information for the passenger, arguably.

app/views/passengers/_form.html.haml
```diff
@@ -15,17 +15,14 @@
         = form.form_group :registration_status,
           label: { text: 'Registration Status' },
           help: "You can select 'pending' if the passenger is temporary but has no verifying information yet" do
-          = form.radio_button :active_status, 'pending', class: 'form-check-input'
-          = form.radio_button :active_status, 'active', class: 'form-check-input'
-          = form.radio_button :active_status, 'archived', class: 'form-check-input'
+          = form.radio_button :registration_status, 'pending', class: 'form-check-input'
+          = form.radio_button :registration_status, 'active', class: 'form-check-input'
+          = form.radio_button :registration_status, 'archived', class: 'form-check-input'
```
Anbranin, line 20: What are these changes doing here?

app/views/passengers/_form.html.haml
```diff
@@ -15,17 +15,14 @@
...
 
       = form.select :mobility_device_id,
         options_from_collection_for_select(MobilityDevice.order(:name),
           :id, :name, passenger.mobility_device&.id), include_blank: 'None'
 
-      = form.form_group do
-        = form.check_box :permanent, disabled: !@current_user&.admin?
-
```
Anbranin, line 28: So... now no one at all can edit the `permanent` status of a passenger. That's not what we want.

Anbranin (changes requested):
I don't think you ran tests, please remember to do that any time you make a PR.

app/views/passengers/_form.html.haml
jpoleary: I must have branched from my last branch instead of master (or accidentally made some changes in master as well as that branch).

Anbranin: Remember `git add -i` and `git add --patch` are your friends. Even `git diff`

Anbranin (approved):

PR #291, opened 2021-07-20
validation will not run properly on update

Anbranin:
requires_verification? returns false before the call to update.
must set desired status before the SQL insert statement is attempted.

I was able to activate the registration of WHOMEVER, without them having entered the required info. Tricky!


jpoleary:
Would it be worth it to throw an exception if there's an issue saving/validating the passenger or returning an indicator? Doesn't really seem super necessary at the moment but might make hunting bugs down in the future.

jpoleary (approved):

Anbranin:
> Would it be worth it to throw an exception if there's an issue saving/validating the passenger or returning an indicator?

That's already covered by the way the controller is set up. The errors are returned in the flash:
```
flash[:danger] = @passenger.errors.full_messages
```

PR #342, opened 2023-02-07
Use a GitHub OAuth app for feedback submission

### LINKED ISSUE #337: "Site Feedback" inop
Looks like the production token is invalid.

I can do a quick-fix — well, actually it would have to be @sherson because 2FA — but it also looks like this app was set up with a "Personal Access Token" belonging to @umasstransit (Look at me using the passive voice as if it wasn't me who did that). GitHub would really like us to _not_ do that, and PATs can't have no expiration date anymore.

It should probably be set up to be GitHub app, with one authorized user (like the Avail Log on umasstransit.org).

werebus:
Rather than use @umasstransit's personal access token.

Closes #337


config/deploy.rb
```diff
@@ -11,7 +11,7 @@
 
 append :linked_files,
        'config/database.yml',
-       'config/application.yml'
+       'config/credentials/production.key'
```
werebus, line 14: Production key is on the server.

config/credentials/production.yml.enc
```diff
@@ -0,0 +1 @@
+hP5ClNN6t62jeWZOvILFTGYXzwlr6Q+o1uK8ZBVxZaf26Q39/fKBagAiOJu5E23aKAb0HQxi+ewPoGJoStLnFlJ2wWr8mIEMrIk9H243YKwsbji7M2U8HEApWHRyJXzmdqrssaUuuRSQAs6ilue/Fe4G1l4LJaw7EfFW7g==--cT02rktxlmprEp4+--ZuYO6Hq5Z42cABinDmX7pQ==
```
werebus, line 1: Created a new GitHub app, credentials are here.

The production key is in the password database.

app/models/issue_token.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class IssueToken < ApplicationRecord
```
benmelz, line 3: I might question, why do we need to store a single global token in the database rather than just putting it in the credentials file?

dirschn: > I might question

As a matter of fact, you did indeed question

werebus: The only way to _get_ the token is through the web callback flow.

Some other options, I guess:

* Figure out how to edit the encrypted credentials from within the controller action (seems dangerous)
* `File.write` the value to some _other_ configuration file.
* Just output the token value in the rendered view and make Adam copy-and-paste that to the production credentials file.

benmelz: Oh I see, I've worked with these tokens in development (specifically, my own from the round-three production database) but haven't looked into the actual generation process. I guess assumed you could somehow generate them directly on github.

Expanding on the third option, we might be able to do the generation process entirely in a development environment?

I guess the only point of doing all that would be to get rid of chunks of code that you've already written though.

werebus: Yeah, and it makes it into a thing that is... within @sherson's capabilities, but he'll probably need help getting the app set up and a walk through on encrypted credential editing, etc. As is, he just has to click a "yes" button.

benmelz (approved):

PR #352, opened 2023-04-24
Some small Brochure layout changes

werebus:
1. Wrap everything in a `.container` - to keep it from going to wide.
2. Allow two tables to sit next to each-other on lg+ screens
3. Place table caption at the top and make it more prominent; more like a "title".

### DIFF
--- a/app/views/passengers/brochure.html.haml
+++ b/app/views/passengers/brochure.html.haml
@@ -1,76 +1,75 @@
-%h1 SpecTrans Passenger Information
-.row
-  .col-md-6
-    %table.table
-      %caption Contact Information
-      %thead
-      %tbody
-        %tr
-          %th Phone
-          %td 413-545-2086
-        %tr
-          %th Public Website
-          %td= link_to 'www.umass.edu/transportation/special-transportation',
-            'https://www.umass.edu/transportation/special-transportation'
-        %tr
-          %th Email
-          %td stspv@umass.edu
-.row
-  .col-md-6
-    %table.table
-      %caption Special Transportation Staff
-      %thead
-      %tbody
-        %tr
-          %th Supervisors
-          %td Lisa Gagnon &amp; Shayla Keane
-        %tr
-          %th Coordinator
-          %td Diana Noble
-%h2 Hours of Operation
-.row
-  .col-md-6
-    %table.table
-      %caption Semester
-      %thead
-        %tr
-          %th{ 'aria-label': 'Days' }
-          %th Dispatch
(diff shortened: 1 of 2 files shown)


app/views/passengers/brochure.html.haml
```diff
@@ -1,76 +1,75 @@
...
+            %th Email
+            %td stspv@umass.edu
+    .col-lg-6
+      %table.table
+        %caption Special Transportation Staff
+        %thead
+        %tbody
+          %tr
+            %th Supervisors
+            %td Lisa Gagnon &amp; Shayla Keane
```
dirschn, line 26: Not a layout change, but these are not the current supervisors anymore. Should be Aurora O'Connor (maybe Rory, I know they tend to prefer that name), Karina Patterson (outgoing) and Zack Suranofsky (incoming).

app/views/passengers/brochure.html.haml
```diff
@@ -1,76 +1,75 @@
...
+            %th Coordinator
+            %td Diana Noble
+  %h2 Hours of Operation
+  .row
+    .col-lg-6
+      %table.table
+        %caption Semester
+        %thead
+          %tr
+            %th{ 'aria-label': 'Days' }
```
dirschn, line 37: Again, not necessarily layout, more just a curiosity, but why not just make this normal header text? seems odd to me.

dirschn (approved):

app/views/passengers/brochure.html.haml
werebus: Yes, I have a different PR pending with _content_ updates. Primarily ST -> AVS, but I also intend to ask Aurora to once-over the brochure.

dirschn: I didn't see the PR in my notifs, I'll look through the repo.

app/views/passengers/brochure.html.haml
werebus: I don't remember, but I'm pretty sure this page was initially constructed by copying from a Word document, so it was a decision made by some staff member, copied to Haml by @Anbranin, and then she decided to add a little bit of A11y help without changing the formatting.

Practically speaking, I think the content in this column is pretty evident... once you get to it. But this extra hint is possibly helpful for screen-reader users?

dirschn: Yeah makes sense to me, the only reason I actually noticed it was because the tables look a little funny with no header on the left column. But it doesn't really matter

Issue #328, opened 2022-11-28
Change ST to AVS


werebus:
The ST department has changed its name to "Accessible Van Service". Instances of "Special Transportation" should be changes to "Accessible Van Service", instances of "ST" should be changed to "AVS"

I don't know if internal class names, configuration keys, etc. really need to be changed. However, the domain name of the production server should be, which means the following non-code tasks also need to be done:

* [ ] Add `CNAME`
* [ ] Expand SSL cert

werebus:
* [ ] Get Shibboleth working on new domain

PR #354, opened 2023-04-27
Replace ST with AVS

werebus:
I went the extra step of moving the department, staff, and contact info into the locales file.

This is almost all of #328, but I'm still working on redoing auth on this app in order to allow us to deploy the app at a different URL than st-pax.admin.umass.edu.

### DIFF
--- a/app/controllers/passengers_controller.rb
+++ b/app/controllers/passengers_controller.rb
@@ -42,7 +42,7 @@ class PassengersController < ApplicationController
 
   def edit
     if @registrant.present? && !@registrant&.pending?
-      flash[:warning] = 'To edit your profile, please call (413) 545-2086'
+      flash[:warning] = "To edit your profile, please call #{I18n.t 'department.phone'}"
       redirect_to passenger_path(@registrant)
     end
     @verification = @passenger.eligibility_verification || EligibilityVerification.new
--- a/app/controllers/passengers_controller.rb
+++ b/app/controllers/passengers_controller.rb
@@ -42,7 +42,7 @@ class PassengersController < ApplicationController
 
   def edit
     if @registrant.present? && !@registrant&.pending?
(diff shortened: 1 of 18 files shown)


app/views/sessions/unauthenticated.haml
```diff
@@ -1,20 +1,17 @@
...
-    %li
-      You are an employee of UMass Transit, but we haven't added you to our
-      database yet. In this case, please talk to an ST Supervisor or contact
-      IT.
+      = succeed '.' do
+        = link_to 'brochure', brochure_passengers_path
     %li
-      You are an employee of UMass Transit and should be able to log in.
-      If this is the case, please talk to an ST Supervisor to get it sorted
-      out.
```
werebus, line 17: I couldn't see much meaningful difference between these two scenarios. They're both, "you work here, but don't have access, ask staff."

config/locales/en.yml
```diff
@@ -60,3 +60,15 @@ en:
       user:
         active: 'Active?'
         admin:  'Admin?'
+  department:
+    name: Accessible Van Service
+    abbr: AVS
+    phone: 413-545-2086
+    email: stspv@umass.edu
```
werebus, line 67: They don't have a new email address yet.

app/controllers/passengers_controller.rb
```diff
@@ -42,7 +42,7 @@ def new
 
   def edit
     if @registrant.present? && !@registrant&.pending?
-      flash[:warning] = 'To edit your profile, please call (413) 545-2086'
+      flash[:warning] = "To edit your profile, please call #{I18n.t 'department.phone'}"
```
benmelz, line 45:
```suggestion
      flash[:warning] = "To edit your profile, please call #{t 'department.phone'}"
```
`t` helper should be mixed in here I think?

benmelz (approved):

PR #355, opened 2023-04-27
Use layout for shared content between mailer views

werebus:
I also spit the `.haml` views into `.html.haml` and `.text.erb` views. And, I added a mailer preview.


config/application.rb
```diff
@@ -30,5 +30,7 @@ class Application < Rails::Application
     # in config/environments, which are processed later.
 
     config.time_zone = 'Eastern Time (US & Canada)'
+
+    config.action_mailer.preview_path = Rails.root.join('app/mailers/previews')
```
benmelz, line 34: I like this location, we should apply this to other apps as well

benmelz (approved):

config/application.rb
werebus: Turns out Zeitwerk has a complaint, though...

PR #359, opened 2023-06-15
Properly name LogEntry system components

### LINKED ISSUE #358: View template reloading doesn't work for `app/views/log/*`
@aflam745 discovered while working on #336 that when editing `app/views/log/index.haml`, any changes made aren't reflected until a full server restart.

On a hunch, I thought that perhaps `log/` has a special rule that makes it ignored by the rails' file-watcher-code-reloading system aaaand when I rename the folder in development, things start working again.

`log` is a bad name for this controller/view system anyways - the model is named `LogEntry`.

benmelz:
Closes #358

I preserved the front facing path (`/log`) but now everything is properly resourceful behind the scenes


config/routes.rb
```diff
@@ -6,7 +6,7 @@
   get '/github/callback' => 'github#callback'
 
   resources :feedback, only: %i[index show new create]
-  resources :log, except: %i[edit new show]
+  resources :log_entries, only: %i[index create update destroy], path: '/log'
```
werebus, line 9:
```suggestion
  resources :log_entries, only: %i[index create update destroy], path: 'log'
```

(At least according to the example in the Rails guide; seems to work the same either way)

app/views/log_entries/index.haml
```diff
@@ -19,11 +19,11 @@
           - if @current_user.can_modify? entry
             .edit-form
               = bootstrap_form_with namespace: entry.id, model: entry,
-                url: log_path(entry), local: true do |f|
+                url: log_entry_path(entry), local: true do |f|
```
werebus, line 22: We can just remove the `url:`. If you give `form_with` a `model:`, it will infer the path.

app/views/log_entries/index.haml
```diff
@@ -19,11 +19,11 @@
...
                 = f.text_area :text, required: true, rows: 8, hide_label: true
                 = f.primary 'Change log entry'
         .actions.col-xl-1.form-inline
           - if @current_user.can_modify? entry
             = link_to 'Edit', '#', class: 'edit-entry mr-5'
-            = button_to 'Delete', log_url(entry), method: :delete,
+            = button_to 'Delete', log_entry_path(entry), method: :delete,
```
werebus, line 28: Similar here:

```suggestion
            = button_to 'Delete', entry, method: :delete,
```

werebus (changes requested):

PR #368, opened 2023-06-21
Add LogEntry request specs

benmelz:
We just overhauled these and I noticed they're completely uncovered. So here's some comprehensive request specs.


spec/requests/log_entries_spec.rb
```diff
@@ -0,0 +1,329 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Log Entries' do
```
dirschn, line 5: This is more of a general question than anything, but in round-three we have [Title Case](https://github.com/umts/round-three/blob/9f3e226b1db721a7409775cd9d2532d63c07ac75/spec/requests/job_types_spec.rb#L3), [PascalCase](https://github.com/umts/round-three/blob/9f3e226b1db721a7409775cd9d2532d63c07ac75/spec/requests/assignment_users_spec.rb#L3) and sorta kinda [snake_case](https://github.com/umts/round-three/blob/master/spec/requests/employee_assignments_spec.rb#L3) but also it's just the relative URL... what I'm getting at is, is Title Case what we want to lock in for the scenario?

dirschn (approved):

spec/requests/log_entries_spec.rb
benmelz: That's where I've settled recently yeah

PR #376, opened 2023-09-20
Replace will_paginate with kaminari

### LINKED ISSUE #363: Move from `will_paginate` to `kaminari`
`will_paginate` is no longer maintained, `kaminari` is the way forward

dirschn:
closes #363 

All config and styling was ripped straight from round-three


app/controllers/log_entries_controller.rb
```diff
@@ -28,7 +28,7 @@ def destroy
   def index
     @entry = LogEntry.new
     @entries = LogEntry.includes(:user).order(pinned: :desc, created_at: :desc)
-                       .page(params[:page] || 1)
+                       .page(params[:page] || 1).per(100)
```
dirschn, line 31: This was how it was set before. Can remove if the default 25 is preferred.

app/views/kaminari/_paginator.html.haml
```diff
@@ -0,0 +1,10 @@
+= paginator.render do
+  %nav{ 'aria-label': 'Results pages' }
+    %ul.pagination.justify-content-center
+      = prev_page_tag
+      - each_page do |page|
+        - if page.left_outer? || page.right_outer? || page.inside_window?
+          = page_tag page
+        - elsif !page.was_truncated?
+          = gap_tag
+      = next_page_tag
```
benmelz, line 10: > All config and styling was ripped straight from round-three

Was it? This doesn't reflect @frothedoatmilk and I's "responsive" implementation.

dirschn: Ah, I took it from your original PR, not the current generation. I'll update them.

app/controllers/log_entries_controller.rb
benmelz, line 31:
```suggestion
                       .page(params[:page]).per(100)
```

benmelz (approved):

PR #382, opened 2023-09-22
Ease up on the passenger index

werebus:
Turns out that there are a ton of (invalid) Passengers in prod with a `nil` Spire ID. I'm just going to punt on this one given that I _hope_ to be removing that column at some point.

### DIFF
--- a/db/migrate/20230922123246_add_spire_unique_indices.rb
+++ b/db/migrate/20230922123246_add_spire_unique_indices.rb
@@ -1,6 +1,6 @@
 class AddSpireUniqueIndices < ActiveRecord::Migration[6.1]
   def change
-    add_index :passengers, :spire, unique: true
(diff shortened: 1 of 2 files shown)


db/migrate/20230922123246_add_spire_unique_indices.rb
```diff
@@ -1,6 +1,6 @@
 class AddSpireUniqueIndices < ActiveRecord::Migration[6.1]
   def change
-    add_index :passengers, :spire, unique: true
+    add_index :passengers, :spire
```
werebus, line 3: I'm going to assume that no one has pulled this migration and run it since it hit `main`. Although if they have, it'll just show up as a confusing `schema.rb` difference.

benmelz (approved):

PR #392, opened 2024-01-18
Rubocopped spec/{controllers,libraries,mailers}

werebus:
Now that we have rubocop-rspec, a bunch more offenses.


spec/libraries/mock_github_client_spec.rb
```diff
@@ -18,15 +18,21 @@
...
 
     it 'returns a dummy issue' do
       expect(call).to respond_to(:number, :title, :body, :labels)
+    end
+
+    it 'returns a dummy issue with lables' do
       expect(call.labels).to be_an(Array)
+    end
+
+    it 'returns dummy lables' do
```
dirschn, line 35:
```suggestion
    it 'returns a dummy issue with labels' do
      expect(call.labels).to be_an(Array)
    end

    it 'returns dummy labels' do
```

dirschn (changes requested):

dirschn (approved):

PR #393, opened 2024-01-18
Rubocopped model specs

werebus:


spec/models/passenger_spec.rb
```diff
@@ -3,125 +3,161 @@
...
-    it 'correct email is sent upon creation of active passenger' do
-      mail = ActionMailer::MessageDelivery.new(PassengerMailer, :notify_active)
-      expect(PassengerMailer).to receive(:notify_active).and_return mail
-      expect(PassengerMailer).not_to receive(:notify_pending)
-      expect(PassengerMailer).not_to receive(:notify_archived)
-      expect(mail).to receive(:deliver_now).and_return true
+  let(:passenger) { create(:temporary_passenger, :with_note) }
+
+  before do
+    @passenger = create(:temporary_passenger, :with_note)
```
dirschn, line 9: is this `before` needed at all? definitely not the instance variable

spec/models/passenger_spec.rb
```diff
@@ -3,125 +3,161 @@
...
-      expect(PassengerMailer).to receive(:notify_pending).and_return mail
-      expect(PassengerMailer).not_to receive(:notify_active)
-      expect(PassengerMailer).not_to receive(:notify_archived)
-      expect(mail).to receive(:deliver_now).and_return true
-      create(:passenger, registration_status: 'pending')
+    it 'sends the correct email' do
+      expect(PassengerMailer).to have_received(:notify_active)
+    end
+
+    it 'does now sent the incorrect email' do
```
dirschn, line 25:
```suggestion
    it 'does not sent the incorrect email' do
```

I think?

spec/models/passenger_spec.rb
```diff
@@ -3,125 +3,161 @@
...
-          expect { @passenger.set_status('pending') }
-            .to(change { @passenger.registration_status })
-        end
-
-        it 'sets the status to pending' do
-          @passenger.pending!
-          expect { @passenger.set_status('active') }
-            .to(change { @passenger.registration_status })
-        end
+    it 'does now sent the incorrect email' do
```
dirschn, line 49:
```suggestion
    it 'does not sent the incorrect email' do
```

spec/models/passenger_spec.rb
```diff
@@ -3,125 +3,161 @@
...
+
+  describe '#temporary?' do
+    subject(:call) { passenger.temporary? }
+
+    it 'returns true if the passenger is not permanent' do
+      expect(call).to be(true)
+    end
+  end
+
+  describe 'not having a SPIRE' do
```
dirschn, line 114: I get you're just doing the minimum to appease rubocop but this also totally feels like an improper use of `describe`. I don't think I'll include this in requested changes, just wanted to note that in case it itches your brain too

spec/models/passenger_spec.rb
```diff
@@ -3,125 +3,161 @@
...
-            date = 14.days.from_now.to_date
-            @passenger.eligibility_verification.update(expiration_date: date)
-            expect(@passenger.rides_expire).to eq 3.business_days.after(date)
-          end
-        end
+    context "with a non-expired doctor's note" do
+      it 'returns three business days after the expiration date of the note' do
+        date = 14.days.from_now.to_date
+        passenger.eligibility_verification.update(expiration_date: date)
+        expect(call).to eq(3.business_days.after(date))
```
dirschn, line 143: apparently these aren't offenses, but they should also probably make use of `before`s? Again, rubocop is appeased so not requested, just pointing it out

spec/models/passenger_spec.rb
```diff
@@ -3,125 +3,161 @@
...
+
+  describe '#expiration_display' do
+    subject(:call) { passenger.expiration_display }
+
+    it 'returns nil for a permanent' do
+      passenger.update permanent: true
+      expect(call).to be_nil
+    end
+
+    it 'returns the expiration date of the doctors note for temporary passengers' do
```
dirschn, line 100:
```suggestion
    it "returns the expiration date of the doctor's note for temporary passengers" do
```

dirschn (changes requested):

spec/models/passenger_spec.rb
werebus: Yeah, I don't love this spec. But FWIW, I do have a branch that removes it... somewhere in my project to stop using Spire at all on this app.

spec/models/passenger_spec.rb
werebus: 'Tis not. My strategy was to add the `let`, progressively change the usage in the specs, then remove the instance variable. Just forgot that last step.

spec/models/passenger_spec.rb
werebus: Yeah, I'm not sure. I sometimes use `before` blocks when both they "make sense" and Rubocop thinks the example is too long, but I don't know if there's a benefit to that indirection in cases when the "before" steps only apply to the one example?

dirschn (approved):

PR #394, opened 2024-01-18
Rubocopped system specs

werebus:


spec/system/passenger_management_spec.rb
```diff
@@ -3,142 +3,145 @@
...
-      @verifying_agency = create(:verifying_agency)
-      when_current_user_is(@user)
-    end
+  let!(:passenger) { create(:passenger, name: 'Foo Bar') }
+  let!(:verifying_agency) { create(:verifying_agency) }
+
+  context 'when the user is an admin' do
+    before { when_current_user_is :admin }
+
+    context 'when when creating a new passenger' do
```
dirschn, line 12:
```suggestion
    context 'when creating a new passenger' do
```

spec/system/passenger_management_spec.rb
```diff
@@ -3,142 +3,145 @@
...
         end
       end
 
-      context 'passenger creation unsuccessful' do
-        it 'checks for existing passengers if a duplicate spire is found' do
-          visit passengers_path
-          click_on 'Add New Passenger'
-          fill_in 'Spire', with: "#{@passenger.spire}\t"
+      context 'when trying to use a duplicate Spire ID' do
+        before { fill_in 'Spire', with: "#{passenger.spire}\t" }
```
dirschn, line 44: Just curious, is there a reason the tab is needed?

spec/system/passenger_management_spec.rb
```diff
@@ -3,142 +3,145 @@
...
+
+        it 'warns about the duplication' do
           expect(page).to have_text 'A passenger already exists for this Spire ID'
+        end
+
+        it 'Allows it anyways (for some reason?)' do
           expect(page).to have_button 'Add new passenger'
+        end
+
+        it 'Offers a link to edit the existing passenger' do
```
dirschn, line 54:
```suggestion
        it 'allows it anyways (for some reason?)' do
          expect(page).to have_button 'Add new passenger'
        end

        it 'offers a link to edit the existing passenger' do
```

spec/system/passenger_management_spec.rb
```diff
@@ -3,142 +3,145 @@
...
           fill_in 'Name', with: 'Foo Bar'
           fill_in 'Email', with: 'foobar@invalid.com'
           fill_in 'Address', with: '123 turkey lane'
           fill_in 'Phone', with: '123'
           fill_in 'Spire', with: '12345678@umass.edu'
-          fill_in 'How long will the passenger be with us?',
-                  with: 2.days.since.strftime('%Y-%m-%d')
-          select @verifying_agency.name,
-                 from: 'Which agency verifies that this passenger needs rides?'
+          fill_in 'How long will the passenger be with us?', with: 2.days.since.strftime('%Y-%m-%d')
```
dirschn, line 25:
```suggestion
          fill_in 'How long will the passenger be with us?', with: 2.days.from_now.strftime('%F')
```

You use it later, might as well be consistent. Also improves readability, in my opinion (specifically `from_now`)

spec/system/passenger_registration_spec.rb
```diff
@@ -2,64 +2,88 @@
...
+      let(:passenger) { create(:temporary_passenger, :with_note) }
+
+      it 'redirects to the show page' do
+        expect(page).to have_current_path passenger_path(passenger)
+      end
+
+      it 'redirects to the show page for registering' do
         visit register_passengers_path
+        expect(page).to have_current_path passenger_path(passenger)
+      end
```
dirschn, line 80: I don't find the name of this particularly intuitive... it took me a little while to figure out what is going on here. 

maybe this?
```suggestion
      it 'redirects to the show page instead of registering' do
        visit register_passengers_path
        expect(page).to have_current_path passenger_path(passenger)
      end
```

spec/system/passenger_registration_spec.rb
```diff
@@ -2,64 +2,88 @@
...
+
+      it 'does not allow editing' do
         expect(page).to have_no_field 'Address'
```
dirschn, line 83: I assume field means input? Because there certainly is an address "thing" on the show page... I think I have a problem overall with the names of the examples as someone who hasn't really seen this site or these specs before, but I don't know if that matters much for this PR

spec/system/user_management_spec.rb
```diff
@@ -3,77 +3,70 @@
...
     end
-  end
 
-  context 'deleting a user successfully', :js do
-    it 'deletes the user and says it did' do
-      visit users_path
-      page.accept_confirm 'Are you sure?' do
-        within('tr', text: @user.name) { click_on 'Destroy' }
+    context 'when a user is deletable' do
+      it 'deletes the user and says it did' do
```
dirschn, line 30: I don't see a test for whether it actually deletes the user.... then again later it's got the same kind of implications. Is that just because it's a system test, don't bother testing that it actually updates that's for request/model specs?
```suggestion
      it 'says it deletes the user' do
```

spec/system/user_management_spec.rb
```diff
@@ -3,77 +3,70 @@
...
-  context 'editing an existing user unsuccessfully' do
     it 'puts the error in the flash' do
-      visit users_path
-      within('tr', text: @user.name) { click_on 'Edit' }
       fill_in 'Spire', with: 'not a valid spire'
-      click_button 'Save'
-      expect(page).to have_text 'Spire must be 8 digits followed by @umass.edu'
+      click_on 'Save'
+      expect(page).to have_text('Spire must be 8 digits followed by @umass.edu')
     end
```
dirschn, line 52: Did you drop the successful/unsuccessful contexts on purpose?

dirschn (changes requested):
I think I might be a little more... intense on these than perhaps was realistically desired. If the intention of these is purely to fix all the rubocop errors and not necessarily make the tests be good, then I'll back off a little

spec/system/passenger_registration_spec.rb
werebus: Yeah, that's a Capybara convention. "field" means `input` or `textarea`.

spec/system/user_management_spec.rb
werebus: Same for all of these examples, actually.

spec/system/user_management_spec.rb
werebus: Yeah, I think context descriptions like these don't make sense. They don't describe "context", they describe the examples they contain. They were also empty except for the one example they each contained.

spec/system/user_management_spec.rb
dirschn: > Same for all of these examples, actually.

Yeah i noticed that after and that's when I started thinking I might be going a little too hard considering the state of these specs

spec/system/user_management_spec.rb
dirschn: Cool, just wanted to check because seeing set up stuff inside an example sets off bells in my brain.

spec/system/passenger_management_spec.rb
```diff
@@ -148,56 +151,52 @@
...
   end
 
-  context 'as a dispatcher' do
-    before do
-      @user = create(:user)
-      @passenger = create(:passenger, name: 'Foo Bar')
-      when_current_user_is(@user)
-    end
+  context 'when the user is a dispatcher' do
+    before { when_current_user_is :anyone }
```
dirschn, line 155: Here's another rabbit hole, I guess users are either dispatchers or admin. fun stuff, but it's fine for now

spec/system/passenger_management_spec.rb
dirschn, line 1: This file has a bunch more "it does this but I don't actually check that I only look for the flash"

spec/system/passenger_management_spec.rb
werebus: It being a dispatch-only tool, the base type of `User` is a dispatcher. This app is a little weird in that you can log in as a `Passenger`, too. The helper method also accepts `:anybody`, I could add `:dispatcher` as another synonym, but I wonder if that would misleadingly imply some sort of special permissions?

dirschn: That's what I wasn't sure about, whether or not passengers could log in. Unless it's something you want to fix particularly bad I'm okay with this staying as is based on [this tidbit](https://github.com/umts/st-pax/blob/19bbdae20664f89fb61abee5508c3b2dbd26e762/app/models/user.rb#L11)

dirschn (approved):
Thanks for humoring me

PR #396, opened 2024-01-22
Don't use instance variables in helpers

werebus:
At least partially by just not using helpers.


app/views/passengers/_form.html.haml
```diff
@@ -36,15 +36,15 @@
             options_for_select(VerifyingAgency.all.map { |agency| [agency.name,
             agency.id, { 'data-needs-contact-info': agency.needs_contact_info? } ] },
             passenger.eligibility_verification&.verifying_agency&.id),
-            help: verifying_agency_help_text,
-            label: verifying_agency_label,
+            help: 'Required only for temporary passengers.',
```
dirschn, line 39: I mean.... while you're here... changing things... might be good to lean into using locale texts? Looks like this app already uses them to a greater extent than round-three

werebus: :thinking: perhaps.

PR #397, opened 2024-01-22
Move message strings into locale settings

werebus:


app/models/eligibility_verification.rb
```diff
@@ -9,11 +9,8 @@ class EligibilityVerification < ApplicationRecord
   validates :passenger, uniqueness: true
   validates :expiration_date, presence: true, if: :passenger_requires_validation
   validates :verifying_agency_id, presence: true, if: :passenger_requires_validation
-  validates :expiration_date, absence: { if: -> { verifying_agency.blank? },
-                                         message: 'cannot be entered without verifying agency' }
-  validates :expiration_date,
-            absence: { if: -> { passenger&.permanent? },
-                       message: 'may not be entered for permanent passengers.' }
+  validates :expiration_date, absence: { if: -> { verifying_agency.blank? }, message: :no_agency }
+  validates :expiration_date, absence: { if: -> { passenger&.permanent? }, message: :perm_pax }
```
werebus, line 13: I was a little mad about this; I figured out this usage by just guessing it would work. The [Rails Guide](https://guides.rubyonrails.org/active_record_validations.html#message) says:

> The `:message` option accepts either a `String` or `Proc` as its value.

dirschn: Yeah I had to do validation messages a while ago and it took me way too long to figure out how to set it up

config/locales/en.yml
```diff
@@ -76,3 +78,38 @@ en:
       - "Aurora O'Connor"
       - Karina Patterson
       - Zack Suranofsky
```
dirschn, line 80: Sidenote, this is outdated quite a bit now

config/locales/en.yml
dirschn, line 1: We've been not using periods elsewhere for a while now, mostly because default rails messages (`errors.full_messages.to_sentence`) don't have periods

dirschn (changes requested):

config/locales/en.yml
werebus: I''ll do another PR, I can't remember whether @sherson still has a "things to do when a new staff member is hired" document, but this should go on there.

dirschn (approved):

PR #400, opened 2024-01-26
Store AVS staff titles in the database

werebus:
Was updating the AVS supervisors which we had flaked on for a bit (#399) and decided it might be better to put that bit of information on the `User` model, that way AVS can manage it themselves.


dirschn (commented):
The only reservation I have about this is that a simple typo would cause duplicate entries for the same "title" (even simply adding a space at the end by accident). Would you consider adding a datalist that pulls from all current titles to _suggest_ what they should type in?

werebus:
> The only reservation I have about this is that a simple typo would cause duplicate entries for the same "title" (even simply adding a space at the end by accident). Would you consider adding a datalist that pulls from all current titles to _suggest_ what they should type in?

That sounds like a good idea, lightweight enough in that we don't have to maintain their title list for them, but still a little hand-holding.

spec/system/passenger_brochure_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+RSpec.describe 'Passenger Brochure' do
+  let!(:staff_member) { create(:user, title: 'Admiral') }
+  let!(:non_staff_member) { create(:user) }
+  let!(:inactive_staff_member) { create(:user, active: false, title: 'Grand Poobah') }
+
+  before { visit brochure_passengers_path }
+
+  it 'shows users with titles' do
+    expect(page).to have_css('tr', text: /Admiral.*#{staff_member.name}/)
```
dirschn, line 13: Does this ignore all whitespace between 'Admiral' and the name?

werebus: Yes, within the bounds of one `<tr>`. But, I think I want to achieve that slightly differently.

app/views/users/_form.html.haml
```diff
@@ -5,6 +5,9 @@
       = f.text_field :spire, placeholder: '12345678@umass.edu'
       = f.form_group do
         = f.check_box :active
-      =f.form_group do
+      = f.form_group do
         = f.check_box :admin
+      = f.text_field :title, help: 'Will be shown on brochure', list: 'title-list'
```
dirschn, line 10:
```suggestion
      = f.text_field :title, help: 'Will be shown on brochure', list: 'title-list', autocomplete: 'off'
```
I generally prefer turning autocomplete off for `datalist`s, because it'll suggest "bad" data if you made a mistake in the past, but I don't think I'll require this here. Up to you.

app/views/users/_form.html.haml
```diff
@@ -5,6 +5,9 @@
...
+      %datalist#title-list
+        = options_for_select @titles
```
dirschn, line 12: I might suggest this to cut down the extra fluff in the controller, but I think this is another thing I don't have a strong opinion on in this case.
```suggestion
        = options_for_select User.select(:title).distinct.pluck(:title).compact
```

dirschn (approved):

app/views/users/_form.html.haml
werebus: Funny enough, I was also having an issue where LastPass was offering to fill in "title" with my "title" ("Mr").

app/views/users/_form.html.haml
werebus: Went back and forth on this, but I think you're right (as a side effect, also resolves a "don't use instance variables in partials" Rubocop offense).

app/views/users/_form.html.haml
werebus: Of course... Lastpass doesn't believe in `autocomplete='off'`.

dirschn (approved):

PR #401, opened 2024-01-29
Check indexes rather than the database for changes

werebus:
I had second thoughts about some of the specs I added in #394 

I think it's slightly more proper to make assertions about object changes the same way a "real user" would — by checking the website.


benmelz (approved):
My kind of system spec

dirschn (commented):
Damn I didn't think anyone else would get in, I had some comments

PR #406, opened 2024-05-08
Export CSV button added

### LINKED ISSUE #404: CSV of Name, Longer rides, Permanent
Dispatchers need a CSV export with these columns for _Active_ passengers:
 
* Name
* Longer rides
* Permanent

They'd like an `Export` button next to `Copy Emails`.

Replaces #384

aarinmehta:
closes #404


app/controllers/passengers_controller.rb
```diff
@@ -42,6 +42,12 @@ def index
     respond_to do |format|
       format.html
       format.pdf { passenger_pdf }
+      format.csv do
```
werebus, line 45: This seems like the right approach — use the selected filter on the index — but the link you added doesn't use this.

app/controllers/passengers_controller.rb
```diff
@@ -129,4 +135,12 @@ def try_notifying_passenger(success:, failure:, success_message:)
     flash[:warning] = t('.notify_warning', message: success_message)
     success.call
   end
+  
+  def export_csv
```
werebus, line 139: I don't think this route is needed (if my other suggestions are adopted).

app/views/passengers/_filters_and_key.html.haml
```diff
@@ -12,6 +12,7 @@
       = f.submit 'Print'
       -if @current_user.admin?
         %input.btn.btn-secondary#copybtn{type: 'button', value: 'Copy Emails'}
+        = link_to('Export CSV', passengers_path(format: 'csv', status: params[:status]), class: 'btn btn-secondary')
```
werebus, line 15: Can I perhaps point you to [this comment](https://github.com/umts/st-pax/issues/404#issuecomment-2023941484) as a hint? A form can have more than one submit button — in this case one for PDF, one for CSV.

werebus (changes requested):
Also, the Travis failure is due to a coverage drop. If you add and commit your `.coverage/last_run.json` that will signal, "yes coverage dropped and we, as an organization, are OK with that." But, probably better to write some tests for the code you added.

app/views/passengers/_filters_and_key.html.haml
```diff
@@ -9,9 +9,10 @@
               label: filter.capitalize,
               checked: @filter == filter,
               inline: true
-      = f.submit 'Print'
+      = f.submit 'Print', formaction: passengers_path(format: :pdf)
       -if @current_user.admin?
         %input.btn.btn-secondary#copybtn{type: 'button', value: 'Copy Emails'}
+        = f.submit 'Export CSV', formaction: passengers_path(format: :csv), class: 'btn btn-secondary'
```
werebus, line 15: We've got the `bootstrap_form_tag` helper here in this app, it takes care of the `btn-` classes unless your're doing something "different".

```suggestion
        = f.submit 'Export CSV', formaction: passengers_path(format: :csv)
```

werebus (changes requested):

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "line": 87.08
+    "line": 86.43
```
werebus, line 3: So, this fixes the Travis failure, but I would prefer not to have a coverage drop.

A simple test or two of `Passenger.to_csv` should actually take care of it?

werebus (changes requested):

app/models/passenger.rb
```diff
@@ -3,6 +3,8 @@
 require 'csv'
 
 class Passenger < ApplicationRecord
+  attr_writer :needs_longer_rides
+
```
benmelz, line 7:
```suggestion
```
Remember we talked about this:

https://github.com/umts/st-pax/blob/324526a2af2eacb5be376edee8db2621a8d8971a/app/models/passenger.rb#L44-L46

Namely that `needs_longer_rides` is not an attribute on `Passenger`. It's just a method that checks if the passenger has a `MobilityDevice` that has `mobility_device.needs_longer_rides == true`.

As such, we shouldn't be adding anything to the model, you just need to set up your test data correctly - more on that below

spec/models/passenger_spec.rb
```diff
@@ -156,5 +156,22 @@
         expect(call).to eq(3.business_days.after(Time.zone.today))
       end
     end
+
+    describe '.to_csv' do
+      subject(:call) { described_class.all.to_csv }
+
+      before do
+        create(:passenger, name: 'John Doe', needs_longer_rides: true, permanent: true)
+        create(:passenger, name: 'Jane Smith', needs_longer_rides: false, permanent: false)
```
benmelz, line 165:
```suggestion
        create(:passenger, name: 'John Doe', permanent: true)
        create(:passenger, name: 'Jane Smith', permanent: false)
```
Not passenger attributes

benmelz:
```suggestion
        longer_rides_mobility_device = create(:mobility_device, needs_longer_rides: true)
        create(:passenger, name: 'John Doe', mobility_device: longer_rides_mobility_device, permanent: true)
        create(:passenger, name: 'Jane Smith', permanent: false)
```
Make one of the passengers need longer rides, by giving them a mobility device that needs longer rides

spec/models/passenger_spec.rb
```diff
@@ -156,5 +156,22 @@
...
+      end
+
+      it 'formats the passengers as a csv' do
+        expect(call).to eq(<<~CSV)
+          Name,Longer rides,Permanent
+          John Doe,false,true
```
benmelz, line 171:
```suggestion
          John Doe,true,true
```

benmelz (changes requested):

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "line": 87.08
+    "line": 15.83
```
werebus, line 3: Needs a whole test-suite run.

werebus (changes requested):

werebus (approved):

PR #409, opened 2024-05-31
New location of single-logout on Cirus Bridge

werebus:
When UMIT switches over their Shibboleth IdP, the "single-logout" location changes.

Don't merge until said switch-over is scheduled. It's currently set for June 12, but I have a feeling it'll be delayed.


benmelz (approved):

werebus:
New switch-over date is July 17

PR #429, opened 2024-09-23
Add dependabot config

werebus:
And unpinned nearly all gems


Gemfile
```diff
@@ -20,7 +20,7 @@ gem 'rails', '~> 7.0.8'
 gem 'rake'
 gem 'sassc-rails'
 gem 'sprockets-rails'
-gem 'turbolinks', '~> 5'
```
werebus, line 23: "`turbolinks`" is deprecated and archived. There won't be any release newer than the version we have (5.2.1).

We should probably migrate to Turbo at some point

Issue #364, opened 2023-06-15
Move to `cssbundling`


benmelz:
We're currently on sprockets and using `sassc-rails` to build our sass. Should be easy to migrate to and `npm` based solution with `cssbundling`.

PR #447, opened 2024-09-25
Switch from sassc to dartsass

werebus:
~~Note that `sassc` and `sassc-rails` remain in the `Gemfile.lock` because they're dependencies of the Bootstrap (4) gem. Here's a peek into the future :crystal_ball:, that's why I'm doing this. Bootstrap 5 removes that dependency.~~  Edit: decided to go all the way to cssbundling, which required I use Bootstrap from npm.

I also opted for the run-pty based dev script rather than the Foreman one, because I just like it better.

This _might_ need some CI tweaks, too; We'll see in a minute...


werebus:
> This might need some CI tweaks, too

Indeed it does

werebus:
Maybe I want to go a bit further as suggested in #364....

PR #487, opened 2024-10-08
Upgrade Bootstrap to 5.x

werebus:


app/views/log_entries/_form.html.haml
```diff
@@ -3,8 +3,6 @@
   .d-flex
     = f.primary log_entry.persisted? ? 'Save' : 'Submit'
     - if @current_user.admin?
-      .input-group.ml-2
+      .input-group.ms-2
         .input-group-text
-          .custom-control.custom-checkbox
-            = f.check_box_without_bootstrap :pinned, class: 'custom-control-input'
-            = f.label :pinned, class: 'custom-control-label'
+          = f.check_box :pinned, multiple: true
```
werebus, line 8: `multiple` is a bit of a hack to stop `bootstrap_form` from [tacking a `mb-3` on the `.form-check`](https://github.com/bootstrap-ruby/bootstrap_form/blob/b71312021e7989dedda0a094ca8ac57d779969c0/lib/bootstrap_form/inputs/check_box.rb#L87). I could also do this with some actual CSS — it has to be more specific, because `mb-3` is `!important` — or, of course... putting it back the way it was.

benmelz (commented):
Weird small issue that I'm observing - 

Screenshot 2024-12-03 at 11 11 43 AM

There's no "middle dot" on selected radio buttons. I checked in both safari and firefox and also have no idea why it's happening. Is it just me?

benmelz (approved):

PR #488, opened 2024-10-08
Switch from turbolinks to turbo

werebus:


app/views/layouts/_navbar.html.haml
```diff
@@ -36,4 +36,4 @@
     %li.nav-item
       = link_to 'Map Project', map_project_url, class: 'nav-link'
   - if authenticated?
-    = link_to 'Logout', destroy_session_path, class: 'nav-link', method: :delete
+    = button_to 'Logout', destroy_session_path, method: :delete, class: 'btn btn-link nav-link', data: { turbo: false }
```
werebus, line 39: Using `turbo_method` would skip the `turbo:before_cache` that tore-down Datatables.

This lead to the fixed header still showing up on other pages after logging out. Since, in production, you gent sent away to webauth, it doesn't really matter that we keep turbo loaded.

PR #493, opened 2024-10-15
Switch from Travis to GH Actions

werebus:
The setup action doesn't run `bin/setup`, because there wasn't actually that much there that I wanted to do (though most was harmless repetition)

MariaDB runs in a Docker service; weirdly I didn't have to do anything to get headless Chrome to work


.travis.yml
```diff
@@ -1,37 +0,0 @@
...
-  - gem update --system
-  - gem install bundler
-install:
-  - bundle config set --local deployment 'true'
-  - bundle install --jobs=3 --retry=3
-  - yarn
-before_script:
-  - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
-  - chmod +x ./cc-test-reporter
-  - ./cc-test-reporter before-build
```
werebus, line 22: I don't think CodeClimate is actually correctly set up currently

benmelz (approved):

PR #514, opened 2024-11-13
Upgrade to rails 7.1

benmelz:


config/environments/development.rb
```diff
@@ -1,15 +1,12 @@
-require 'passenger_login'
 require "active_support/core_ext/integer/time"
 
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
-  config.middleware.use PassengerLogin
```
werebus, line 7: Removing this breaks signing in as a "New Passenger" using the dev-login.

werebus (changes requested):

config/environments/development.rb
```diff
@@ -1,4 +1,3 @@
-require 'passenger_login'
```
werebus, line 1: `/lib` isn't auto-loaded; need this too.

werebus (changes requested):

werebus (approved):

PR #552, opened 2025-01-13
MobilityDevicesController Returns Correct 404 Error

### LINKED ISSUE #551: Mobility device member pages render nothing on 404
I was working on #550 and realized that `MobilityDevicesController` does not return a 404 if a device can't be found. It loads a blank page. Playing around some more I found that `UsersController` and `PassengersController` return the default Rails error page for an `ActiveRecord::RecordNotFound` error. 

Granted, the only way a person is likely to see this is if they change the `id` parameter in the URL, but I feel like having a 404 page, or redirecting to the respective indices would be good.

alocke8181:
Closes #551 

`MobilityDevicesController#set_device` will now return the standard `ActiveRecord::RecordNotFound` error if the mobility device can't be found. This will render the 404 page in development instead of a blank one. This matches the `PassengersController` and `UsersController`.

### DIFF
--- a/app/controllers/mobility_devices_controller.rb
+++ b/app/controllers/mobility_devices_controller.rb
@@ -48,7 +48,7 @@ class MobilityDevicesController < ApplicationController
   private
 
   def set_device
-    @device = MobilityDevice.find params.require(:id)
+    @device = MobilityDevice.find params[:id]
   end
 
   def device_params
--- a/app/controllers/mobility_devices_controller.rb
+++ b/app/controllers/mobility_devices_controller.rb
@@ -48,8 +48,7 @@ class MobilityDevicesController < ApplicationController
   private
 
   def set_device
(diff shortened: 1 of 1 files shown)


app/controllers/mobility_devices_controller.rb
```diff
@@ -48,8 +48,7 @@ def destroy
   private
 
   def set_device
-    @device = MobilityDevice.find_by id: params.require(:id)
-    head :not_found if @device.blank?
+    @device = MobilityDevice.find params.require(:id)
```
benmelz, line 51:
```suggestion
    @device = MobilityDevice.find params[:id]
```
`.find` calls generally don't need `params.require`- it's okay if it comes in blank because `.find` will throw a `RecordNotFound` anyways

benmelz (changes requested):

benmelz (approved):

PR #553, opened 2025-01-13
Converting 401 - Unauthorized to 403 - Forbidden

alocke8181:
Coverting everything that returns at `401 unauthorized` to return a `403 forbidden` instead. 

@benmelz and I talked, and since all users have to go through the authentication process to get to the website it makes more sense for the controllers to return `forbidden` if the user tries to access a resource they're not allowed to. 

This changes the `ApplicationController#deny_access` by having it take a `reason` as a parameter and outputting a `403` page corresponding to the `reason`. Currently there are 3 reasons: 
- A user does not have an account with the website: `'no_account'`
- A user does have an account, but they are logging in with a subsidiary account: `'subsidiary'`
- A user does have an account, but they do not have access to a resource: `'no_access'`

`no_access` is currently the default simply because there are more ways for it to be called than the others.

Existing specs have been updated to now check for `forbidden` statuses instead of `unauthorized`.


benmelz (changes requested):
Talked about this over teams but documenting secondary here:

- shouldn't need to pass an argument to `deny_access`
- public views need to be fully fledged html files (copy structure/styling from the default rails ones like 404.html)

public/403-no-access.html
```diff
@@ -0,0 +1,43 @@
...
+            You are an AVS employee and believe you should have access. 
+            In this case, please contact AVS staff for assistance.
+          </li>
+        </ul>
+      </p>
+      <p>
+        <form method="post" action="/logout">
+          <a href="#" class="logout-link">Logout</a>
+        </form>
+      </p>
```
benmelz, line 38:
```suggestion
      <h1>You do not have access to this page.</h1>
      <p>Please submit feedback if you believe this is a mistake.</p>
      <nav>
        <a href="/">Home</a>
        <a href="/feedback/new">Feedback</a>
        <form method="post" action="/logout">
          <a href="#" class="logout-link">Logout</a>
        </form>
      </nav>
```
Make this a little more concise/add a home and feedback link

benmelz: See below for styling for the nav

public/403-no-account.html
```diff
@@ -0,0 +1,36 @@
...
+            <a href="http://umass.edu/transit/schedules.html">Schedules</a>, and
+            <a href="htt://umass.edu/transit/join_team.html">Employment Information</a>.
+          </li>
+        </ul>
+      </p>
+      <p>
+        <form method="post" action="/logout">
+          <a href="#" class="logout-link">Logout</a>
+        </form>
+      </p>
```
benmelz, line 31:
```suggestion
      <h1>You do not have access to this page.</h1>
      <p>We don't recognize you as an AVS employee or passenger.</p>
      <p>
        You may be looking for our <a href="https://umass.edu/transit">public website</a>, where you can access news,
         schedules and employment information.
      </p>
      <nav>
        <form method="post" action="/logout">
          <a href="#" class="logout-link">Logout</a>
        </form>
      </nav>
```
Similar deal here

config/routes.rb
```diff
@@ -34,4 +34,5 @@
   end
 
   delete 'sessions/destroy', to: 'sessions#destroy', as: :destroy_session
```
benmelz, line 36: I like the change to `post :logout` - could we then remove this mapping (and adjust any places it's used? Probably just the logout button in the main navbar)

public/403-no-access.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>You do not have access to this page (403)</title>
+  <meta name="viewport" content="width=device-width,initial-scale=1">
+  <script src="status_code.js"></script>
```
benmelz, line 6:
```suggestion
  <script src="/status_code.js"></script>
```

public/403-no-account.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>You do not have access to this page (403)</title>
+  <meta name="viewport" content="width=device-width,initial-scale=1">
+  <script src="status_code.js"></script>
```
benmelz, line 6:
```suggestion
  <script src="/status_code.js"></script>
```

benmelz (changes requested):

benmelz (approved):
Looks good to me I think

PR #590, opened 2025-02-24
Update _scheduling.html.haml

ZackSura:
Modified line 12 to include a recently adjusted policy. Off campus rides cannot be more than 10 minutes travel time, in addition to being within 3 miles.


app/views/passengers/brochure/_scheduling.html.haml
```diff
@@ -9,7 +9,7 @@
     on Friday morning to set up your rides for the following week.
   %li
     #{t 'department.name'} vans travel on campus and in close proximity to the
-    University, we do not travel more than 3 miles from Du Bois Library.
+    University, we do not travel more than 3 miles away or more than 10 minutes from the Du Bois Library.
```
werebus, line 12: This technically fits our style guide, but I've been keeping to 80 characters in the brochure because it's so prose-like.

```suggestion
    University, we do not travel more than 3 miles away or more than 10 minutes
    from the Du Bois Library.
```

werebus (changes requested):
@ZackSura, the next step is to address requested changes. In this case, I made a "suggestion", which you can [accept if you like](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes), or make alternative changes or comments.

ZackSura:
Alright, I've committed the suggested change!

werebus (approved):

werebus:
@ZackSura, your PR has been merged. The last step is actually deploying the site to production.

I'll take care of that some time today - I may as well address some other dependency PRs first.

PR #619, opened 2025-03-18
Upgrade Rails to 7.2

werebus:


bin/setup
```diff
@@ -34,4 +34,8 @@ FileUtils.chdir APP_ROOT do
 
   puts "\n== Restarting application server =="
   system! "bin/rails restart"
+
+  # puts "\n== Configuring puma-dev =="
+  # system "ln -nfs #{APP_ROOT} ~/.puma-dev/#{APP_NAME}"
+  # system "curl -Is https://#{APP_NAME}.test/up | head -n 1"
```
dirschn, line 40: Necessary?

dirschn (approved):

bin/setup
werebus: No, but I try to minimize how much our config differs from the generated config. Doing that means that if it's a difference then we (probably) did it intentionally.

FWIW, the Rails team had second thoughts and this comment will be gone in the upgrade to Rails 8.0.

PR #732, opened 2025-09-16
Switch from yarn to npm

benmelz:
Have always meant to go through the apps and do this but the prospect of umts/workflows pushed me to start.

Used a [tool](https://github.com/imsnif/synp) to convert the lockfile this time, which seems to have gone well. Beyond that it's pretty PnP with rails nowadays.


run-pty.json
```diff
@@ -7,6 +7,6 @@
     "defaultStatus": ["⏳", "S"]
   },
   {
-    "command": ["yarn", "build:css", "--watch"]
+    "command": ["npm", "run", "build:css", "--watch"]
```
werebus, line 10:
```suggestion
    "command": ["npm", "run", "build:css", "--", "--watch"]
```

Otherwise it's interpreted as an argument to `npm`

package-lock.json
```diff
@@ -0,0 +1,1263 @@
...
+        "postcss-nesting": "^13.0.2",
+        "postcss-scss": "^4.0.9"
+      },
+      "devDependencies": {
+        "run-pty": "^5.0.0"
+      }
+    },
+    "node_modules/@csstools/postcss-sass": {
+      "version": "5.1.1",
+      "resolved": "https://registry.yarnpkg.com/@csstools/postcss-sass/-/postcss-sass-5.1.1.tgz",
```
werebus, line 32: So, do we want to keep using Yarn's registry? Arguably that ensures that this change is installing _exactly_ the same packages as before, I guess.

package-lock.json
```diff
@@ -0,0 +1,1263 @@
...
+    },
+    "node_modules/@lydell/node-pty": {
+      "version": "1.0.2",
+      "resolved": "https://registry.yarnpkg.com/@lydell/node-pty/-/node-pty-1.0.2.tgz",
+      "integrity": "sha1-f4IN2RfCfvJ8uEbLZoTpBbOR3is= sha512-5UhusYFuH8GDjmF0+WKAHYv6HrB+AhuGBIcSFHOVRLC4aCf1JMTbrnOtZvXKPQV4MeBupZ9EzN6IZbjvjfI1Ew==",
+      "dev": true,
+      "optionalDependencies": {
+        "@lydell/node-pty-darwin-arm64": "1.0.2",
+        "@lydell/node-pty-darwin-x64": "1.0.2",
+        "@lydell/node-pty-linux-x64": "1.0.2",
```
werebus, line 129: The lock file doesn't contain these optional dependencies which caused me, at least, to to be ably to run `run-pty`:

```
/home/matt/projects/st-pax/node_modules/@lydell/node-pty/requireBinary.js:12
            throw new Error(optionalDependencies === undefined
            ^

Error: The @lydell/node-pty package supports your platform (linux-x64), but it could not find the binary package for it: @lydell/node-pty-linux-x64/pty.node

This can happen if you use the "--omit=optional" (or "--no-optional") npm flag.
The "optionalDependencies" package.json feature is used to install the correct
binary executable for your current platform. Remove that flag to use @lydell/node-pty.
```

werebus (changes requested):

PR #753, opened 2025-09-25
Fix packaging issues

frothedoatmilk:
1) We need timecop for our seeds, but somehow it got dropped from the Gemfile?
2) Unpin of stylelint stuff should be reflected in the package lock


werebus (changes requested):
#754

